PR #72, opened 2024-05-20
GPJ: inference evaluator

gpadres:
# Description

Inference evaluator (main and naive attacks)

## How Has This Been Tested?
Tests included in PR


leakpro/synthetic_data_attacks/anonymeter/evaluators/inference_evaluator.py
```diff
@@ -0,0 +1,220 @@
...
+        guesses = self.guesses
+        secrets = self.secrets
+        #Calculate value matches (depending on self.regression)
+        if self.regression:
+            rel_abs_diff = np.abs(guesses - secrets) / (guesses + 1e-12)
+            value_matches = rel_abs_diff <= self.tolerance
+        else:
+            value_matches = guesses == secrets
+        #Calculate nan matches
+        nan_matches = np.logical_and(pd.isna(guesses), pd.isna(secrets))
```
johanos1, line 83: Is there any reason why we even consider nans in this attack? Is it interesting?

leakpro/synthetic_data_attacks/anonymeter/evaluators/inference_evaluator.py
```diff
@@ -0,0 +1,220 @@
...
+        Self : InferenceGuesses
+            Object containing inference guesses and successes (count).
+
+        """
+        #Set guesses and secrets
+        guesses = self.guesses
+        secrets = self.secrets
+        #Calculate value matches (depending on self.regression)
+        if self.regression:
+            rel_abs_diff = np.abs(guesses - secrets) / (guesses + 1e-12)
```
johanos1, line 78: Is this division by guesses correct? I suppose guesses is a dataframe of size n_attacks x #columns in secret? So why not simply take the absolute difference (elementwise) and check if it is below tolerance?

leakpro/synthetic_data_attacks/examples/inference_example.ipynb
johanos1, line 1: There is another tool one should use for jupyter review. I will look it up for next time.

johanos1 (changes requested):
Looks good. Just some comments on the InferenceGuesses.

leakpro/synthetic_data_attacks/anonymeter/evaluators/inference_evaluator.py
gpadres: Guesses has only one column (only 1 inference secret at the time, e.g. shape (n_attacks x 1)), so the division is correct to standardise the measure and tolerance is a pct deviation.

PR #83, opened 2024-06-26
Memorization

### LINKED ISSUE #78: Implement memorization
# Feature
Enable memorization and privacy score metrics

## What Needs to Be Done
Implement memorization and privacy score (from Why train more, Choi et. al.)

## How Can It Be Tested
Run attacks with enhanced perfromance with memorization and privacy score

### LINKED ISSUE #110: Memorization for offline attacks
# Feature
Adapt the memorization boosting for offline attacks

## What Needs to Be Done
Create and implement suitable assumptions for the offline case

## How Can It Be Tested
Run and verify that offline attacks benefits from the method

### LINKED ISSUE #111: Ruff check fails on tests
# Issue
## Problem Description
Running ruff check finds a lot of errors within tests

## Expected Behavior
Tests should not be checked with ruff

## What Needs to be Done
Set tests to be ignored by ruff

Running ruff check with no errors in tests

henrikfo:
# Description

Summary of changes

- Receive model logits a bit faster
- Implemeted Memorization and Privacy-score ( to select most-vulnerable percentile )

## Resolved Issues

- [ ] fixes #78  
- [ ] fixes #110  
- [ ] fixes #111   

## How Has This Been Tested?
- Run LiRA online attack
- Run LiRA offline attack


config/dev_config/cifar10.yaml
fazelehh, line 1: I think it is cleaner if we don't commit these changes in general.

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -3,11 +3,14 @@
 from abc import ABC, abstractmethod
 
 import numpy as np
+import torch
```
fazelehh, line 6: Previously we tried to avoid importing torch like this and just import modules we want. I think we still want that.

config/audit.yaml
```diff
@@ -12,10 +12,12 @@ audit:  # Configurations for auditing
...
-      training_data_fraction: 0.1  # Fraction of the auxilary dataset to use for this attack (in each shadow model training)
-      num_shadow_models: 8 # Number of shadow models to train
-      online: False # perform online or offline attack
-      fixed_variance: True # Use a fixed variance for the whole audit
+      training_data_fraction: 0.5  # Fraction of the auxilary dataset to use for this attack (in each shadow model training)
+      num_shadow_models: 16 # Number of shadow models to train
+      exclude_logit_threshold: 1 # Set a threshold for when to skip a logit based (Useful for unbalance in IN vs OUT samples)
+      online: True # perform online or offline attack
+      memorization: True # Use memorization score to boost performance
+      memorization_threshold: 0.99 # The most vulnerable percentile ( 0.0 for original threshold )
```
johanos1, line 20: So we do not have an automatic way to do this now? Did you write an issue?

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
 
         """
         pass
+
+    def _memorization(self:Self) -> None:
```
johanos1, line 265: I am wondering if we should have this functionality in abstractMIA or if it makes more sense to have it in utils. Any opinion?

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+        """Run memorization score enhancement.
+
+        Memorization enhances the attack performance by only inlucing vulnerable data points
+
+        """
+
+        self.logger.info("Preparing memorization")
+        if not self.online:
+            self.logger.info("Using the offline version of the attack we make some assumptions in the absence of IN-models")
```
johanos1, line 274: be more explicit. "Some assumption"?

johanos1: Also, we dont have anything for offline? Create issue!

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -53,6 +57,7 @@ def __init__(
                 # out_members will start after the last training index and go up to the number of test indices - 1
                 "out_members": np.arange(len(handler.train_indices),len(handler.train_indices)+len(handler.test_indices)),
             }
+            AbstractMIA.skip_indices = np.zeros(len(AbstractMIA.audit_dataset["data"]), dtype=bool)
```
johanos1, line 60: Should this really be shared among all MIAs? Right now we have used it for RMIA and LiRA but they may not use the same number of shadow models and then these indices will not be the same.

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -3,11 +3,14 @@
...
 
 import numpy as np
+import torch
 from torch.utils.data import DataLoader
+from tqdm import tqdm
 
 from leakpro.import_helper import List, Self, Union
 from leakpro.metrics.attack_result import AttackResult
 from leakpro.model import PytorchModel
+from leakpro.signals.signal import ModelLogits, ModelRescaledLogits
```
johanos1, line 13: Not sure we should compute signal values in abstractMIA... I know we talked about it but I think I am having second thoughts.

leakpro/attacks/mia_attacks/abstract_mia.py
johanos1: Basically, we could create a .py file called memorization

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+
+        # Initialize memorization score
+        self.memorization_score = np.zeros(len(logits), dtype=float)
+        logits = self.softmax_logits(logits)
+        if self.online:
+            for i, (logit, mask, label) in tqdm(enumerate(zip(logits, self.in_indices_mask, self.audit_data._labels))):
+                self.memorization_score[i] = np.mean(logit[mask, label]) - np.mean(logit[~mask, label])
+        else:
+            for i, (logit, target_logit, label) in tqdm(enumerate(zip(logits, target_logits, self.audit_data._labels))):
+                self.memorization_score[i] = np.mean(target_logit[label]) - np.mean(logit[~mask, label])
```
johanos1, line 294: Has this been tested? Does it work?

johanos1, line 294: There is no mask in the enumerate...

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+
+        self.privacy_score = self._privacy_score()
+
+        mem_mask, privacy_mask = self.adjust_memorization_mask()
+
+        self.skip_indices = (self.skip_indices | mem_mask) | privacy_mask
+
+    def _privacy_score(self:Self) -> np.ndarray:
+        """Run memorization score enhancement.
```
johanos1, line 303: update description

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+            target_logits = self.target_logits
+        else:
+            logits_function = ModelRescaledLogits()
+            logits = np.swapaxes(logits_function(self.shadow_models, self.audit_data), 0, 1)
+            target_logits = np.swapaxes(logits_function([self.target_model], self.audit_data), 0, 1).squeeze()
+
+        self.logger.info("Calculating privacy score")
+        privacy_score = []
+
+        if len(self.shadow_models) < 64:
```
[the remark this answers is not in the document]
johanos1: 64 is a magic number.

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -45,17 +45,24 @@ def _configure_attack(self:Self, configs: dict) -> None:
 
         self.shadow_models = []
         self.num_shadow_models = configs.get("num_shadow_models", 64)
+        self.exclude_logit_threshold = configs.get("exclude_logit_threshold", 1)
 
         self.online = configs.get("online", False)
-        self.fixed_variance = configs.get("fixed_variance", False)
+
+        self.memorization = configs.get("memorization", False)
+        self.memorization_threshold = configs.get("memorization_threshold", 0.5)
```
johanos1, line 53: Why do you set different default values here than in the actual memorization function?

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -112,36 +147,20 @@ def prepare_attack(self:Self)->None:
             count_in_samples = np.count_nonzero(self.in_indices_mask)
             if count_in_samples > 0:
                 self.logger.info(f"Some shadow model(s) contains {count_in_samples} IN samples in total for the model(s)")
-                self.logger.info("This is not an offline attack!")
+                self.logger.info("This is not an true offline attack!")
```
johanos1, line 150: ?

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -112,36 +147,20 @@ def prepare_attack(self:Self)->None:
...
         self.logger.info(f"Calculating the logits for all {self.num_shadow_models} shadow models")
-        self.shadow_models_logits = np.swapaxes(np.array(self.signal(self.shadow_models, self.audit_data)), 0, 1)
+        self.shadow_models_logits = np.swapaxes(self.signal(self.shadow_models, self.audit_data), 0, 1)
 
         # Calculate logits for the target model
         self.logger.info("Calculating the logits for the target model")
-        self.target_logits = np.array(self.signal([self.target_model], self.audit_data)).squeeze()
+        self.target_logits = np.swapaxes(self.signal([self.target_model], self.audit_data), 0, 1).squeeze()
+
+        if self.exclude_logit_threshold > 0:
```
johanos1, line 159: I dont think exclude_logit_threshold should exist. We should not use audit points without shadow models. Or am I missing something?

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -159,7 +178,7 @@ def run_attack(self:Self) -> CombinedMetricResult:
         score = []  # List to hold the computed probability scores for each sample
 
         # If fixed_variance is to be used, calculate it from all logits of shadow models
-        if self.fixed_variance:
+        if len(self.shadow_models) < 64:
```
johanos1, line 181: why 64?

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -194,7 +214,7 @@ def run_attack(self:Self) -> CombinedMetricResult:
         score = np.asarray(score)  # Convert the list of scores to a numpy array
 
         # Generate thresholds based on the range of computed scores for decision boundaries
-        self.thresholds = np.linspace(np.min(score), np.max(score), 1000)
+        self.thresholds = np.linspace(np.nanmin(score), np.nanmax(score), 2000)
```
johanos1, line 217: Dont use nanmin and nanmax! It just hides bugs. If we handle the audit points missing in or out models as in rmia, we dont need this.

johanos1 (changes requested):

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -182,10 +201,11 @@ def run_attack(self:Self) -> CombinedMetricResult:
 
             if self.online:
                 in_mean = np.mean(shadow_models_logits[mask])
-                if not self.fixed_variance:
+                if len(self.shadow_models) >= 64:
```
fazelehh, line 204: Seems like a magic number.

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -45,17 +45,24 @@ def _configure_attack(self:Self, configs: dict) -> None:
...
+        self.memorization_threshold = configs.get("memorization_threshold", 0.5)
+        self.privacy_score_threshold = configs.get("privacy_score_threshold", 1)
 
         self.include_train_data = configs.get("include_train_data", self.online)
         self.include_test_data = configs.get("include_test_data", self.online)
 
         # Define the validation dictionary as: {parameter_name: (parameter, min_value, max_value)}
         validation_dict = {
             "num_shadow_models": (self.num_shadow_models, 1, None),
+            "exclude_logit_threshold": (self.exclude_logit_threshold, 0, int(self.num_shadow_models/2)),
```
fazelehh, line 62: Nice with the verification :)

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+        logits = logits - torch.max(logits, dim=-1, keepdim=True).values
+        logits = torch.exp(logits)
+        logits = logits/torch.sum(logits, dim=-1, keepdim=True)
+        return logits.numpy()
+
+    def adjust_memorization_mask(self:Self) -> None:
+        """Adjust thesholds to achieve the desired percentile most vulnerable datapoints."""
+
+        audit_dataset_len = len(self.target_logits)
+        if audit_dataset_len*(1-self.memorization_threshold) < 30:
```
fazelehh, line 354: maybe using a variable like num_requested_datapoint or something like that instead of a number.

fazelehh (dismissed):
Nice code and great ROC curve :) 
It would be beneficial to visualize the data points in the memorization and privacy score space. If you have such plots, I think the latex file is a good place to upload them.

config/dev_config/cifar10.yaml
henrikfo: I don't see where this comment is placed after I pushed the commit, but I don't think changes to particular config files matters as long as it's only simple ones, like changing integers such as batch_size = 32 -> 64 etc.

leakpro/attacks/mia_attacks/abstract_mia.py
henrikfo: Agreed! Removed this one and the whole memorization function from abstract_mia.py, to its own file!:)

config/audit.yaml
henrikfo: We do! For memorization alone there is 5 possible flags that can be used. Only memorization True/False is left in audit.yaml now. A question is what kind of flags do we wanna set/ should be visible  in the audit.yaml file. Should have some sort of standard for this?

leakpro/attacks/mia_attacks/abstract_mia.py
henrikfo: Done! The memorization process have been move to a file called "boosting.py". Here we can define more boosting techniques later, or have it open for others to do.

leakpro/attacks/mia_attacks/abstract_mia.py
henrikfo: Yes these assuptions should be explained. Should that maybe be done in the overleaf document maybe? And only have the offline implementation done in the file?

We have it for offline now, created an issue and resolved it in this PR

leakpro/attacks/mia_attacks/abstract_mia.py
henrikfo: Since memorization is moved from abstract_mia.py and we have another way to handle 0 IN-sample and 0 OUT-sample cases, this has been removed

leakpro/attacks/mia_attacks/abstract_mia.py
henrikfo: Since memorization is moved from abstract_mia.py, this has been removed

leakpro/attacks/mia_attacks/abstract_mia.py
henrikfo: It has been tested, works great in LiRA actually!:)

henrikfo: Fixed it!

leakpro/attacks/mia_attacks/abstract_mia.py
henrikfo: Fixed!

leakpro/attacks/mia_attacks/lira.py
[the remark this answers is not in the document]
henrikfo: Yes this should be added!

leakpro/attacks/mia_attacks/lira.py
henrikfo: Yes, is now commented for why it is used

leakpro/attacks/mia_attacks/abstract_mia.py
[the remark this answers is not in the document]
henrikfo: Added comment about that!

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+        for (logit, target_logit, mask) in tqdm(zip(logits, target_logits, self.in_indices_mask)):
+            in_mean, out_mean = np.mean(logit[mask]), np.mean(logit[~mask])
+
+            if len(self.shadow_models) >= 64:
+                in_std, out_std = np.std(logit[mask]), np.std(logit[mask])
+
+            if self.online:
+                privacy_score.append(np.abs(in_mean-out_mean)/(in_std+out_std+1e-30))
+            else:
+                privacy_score.append(np.abs(target_logit-out_mean)/(2*out_std+1e-30))
```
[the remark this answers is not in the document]
henrikfo: This has been tried and works! the x2 comes from that the online case we add in_std and out_std, and the assumption made for the offline case is that they are roughly the same. thus out_std should be a relatively good replacement for in_std. There might be otherways too.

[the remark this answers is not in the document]
henrikfo: This will more importantly be noted in the overleaf doc.

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+
+        return np.asarray(privacy_score)
+
+    def softmax_logits(self:Self, logits: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
henrikfo: Yes we should, It's not done in this PR though. Will create an issue

leakpro/attacks/mia_attacks/abstract_mia.py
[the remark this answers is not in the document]
henrikfo: Its just an arbitrary number that prevents the user from getting too few datapoints, I thought it was a rule to have there but in not too sure now.. what do you think? Does it make some statistical sense for the FPR/TPR plots?

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+    def adjust_memorization_mask(self:Self) -> None:
+        """Adjust thesholds to achieve the desired percentile most vulnerable datapoints."""
+
+        audit_dataset_len = len(self.target_logits)
+        if audit_dataset_len*(1-self.memorization_threshold) < 30:
+            self.logger.info("Trying to audit <30 datapoints, adjusting to 30 datapoints")
+            self.memorization_threshold = (1-30/audit_dataset_len)
+
+        # Set initial thresholds as given from literature ("why train more...")
+        mem_thrshld = 0.8
```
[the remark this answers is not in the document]
henrikfo: What do you mean?

leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -256,3 +261,117 @@ def run_attack(self:Self) -> Union[AttackResult, List[AttackResult]]:
...
+        priv_thrshld = 2.0
+
+        # If the memorization threshold is set to 0.0, use the initial thresholds
+        if self.memorization_threshold != 0.0:
```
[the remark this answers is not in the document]
henrikfo: The code below does just that, based on the percentile that was choosen, it iterates over the mask(s) until its satisfied

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -45,17 +45,24 @@ def _configure_attack(self:Self, configs: dict) -> None:
 
         self.shadow_models = []
         self.num_shadow_models = configs.get("num_shadow_models", 64)
+        self.exclude_logit_threshold = configs.get("exclude_logit_threshold", 1)
```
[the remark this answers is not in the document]
henrikfo: Removed, using the convention from RMIA

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -84,6 +91,34 @@ def description(self:Self) -> dict:
             "detailed": detailed_str,
         }
 
+    def check_logits(self:Self) -> None:
```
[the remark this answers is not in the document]
henrikfo: Done

leakpro/attacks/mia_attacks/lira.py
henrikfo: To have some default setup, such that we only need to have one flag for memorization. If a user wants to have more customization they can allways add those flags to the config file

leakpro/attacks/mia_attacks/lira.py
henrikfo: Had a bug (when I did the offline attack) where some shadow models had IN-samples. Also a good sanity check so users dont mix up offline attacks with online models.

leakpro/attacks/mia_attacks/lira.py
henrikfo: Removed. Using the convention from RMIA

leakpro/attacks/mia_attacks/lira.py
henrikfo: Added comments why

leakpro/attacks/mia_attacks/lira.py
henrikfo: Done

leakpro/attacks/mia_attacks/abstract_mia.py
johanos1: I think we can have both. Yea, let's write it in the overleaf. Eventually, we also need to have a wiki or something in the github repo but thats for later

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+                self.memorization_score[i] = np.mean(logit[mask, label]) - np.mean(logit[~mask, label])
+        else:
+            for i, (logit, target_logit, label) in tqdm(enumerate(zip(logits, target_logits, self.audit_data_labels)),
+                                                total=len(logits),
+                                                desc="Calculating memorization score",
+                                                leave=False
+                                                ):
+                self.memorization_score[i] = np.mean(target_logit[label]) - np.mean(logit[:, label])
+
+    def _privacy_score(self:Self) -> None:
```
johanos1, line 141: So the privacy score is the same thing as the memorizations core except that we normalize by the std. Does it help to use both or do we get the same performance as with only one of them?

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+        mem_thrshld = 0.8
+        priv_thrshld = 2.0
+
+        # Use the thresholds from the paper
+        if self.memorization_threshold == 0.0:
+            return self.memorization_score > mem_thrshld, self.privacy_score > priv_thrshld
+
+        # Adjust initial thresholds if they are set too high
+        while (np.count_nonzero((self.memorization_score > mem_thrshld)\
+                | (self.privacy_score > priv_thrshld)) < self.num_memorization_audit_points):
```
johanos1, line 203: why or operator, why not and?

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+            mem_thrshld = mem_thrshld/2
```
johanos1, line 204: should these be lowered in sequence rather than simultaneously?

johanos1 (approved):

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+        # Set starting thresholds, from the paper ("why train more...")
+        mem_thrshld = 0.8
+        priv_thrshld = 2.0
+
+        # Use the thresholds from the paper
+        if self.memorization_threshold == 0.0:
+            return self.memorization_score > mem_thrshld, self.privacy_score > priv_thrshld
+
+        # Adjust initial thresholds if they are set too high
+        while (np.count_nonzero((self.memorization_score > mem_thrshld)\
```
johanos1, line 202: The memorization score can be negative. This may put you in a situation where you may never be able to satisfy the condition! I just ran into this bug when working on a different issue.

johanos1 (changes requested):

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+        self.handler = handler
+        self.online = online
+        self.logger = logger
+        self.batch_size = batch_size
+
+    def run(self:Self) -> list:
+        """Run memorization enhancement."""
+
+        self.audit_data_length = len(self.audit_data_indices)
+        if self.org_audit_data_length*(1-self.memorization_threshold) < self.min_num_memorization_audit_points:
```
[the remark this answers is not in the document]
henrikfo: Here we are just checking if we are getting enough points from the choosen threshold, just a sanity check!

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+        Memorization enhances the attack performance by only inlucing vulnerable data points
+        """
+
+        logits_function = ModelLogits()
+        logits = np.swapaxes(logits_function(self.shadow_models, self.handler, self.audit_data_indices, self.batch_size), 0, 1)
+        target_logits = np.swapaxes(logits_function([self.target_model], self.handler, self.audit_data_indices, self.batch_size),\
+                                    0, 1).squeeze()
+
+        logits = self.softmax_logits(logits)
+        target_logits = self.softmax_logits(target_logits)
```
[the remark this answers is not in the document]
henrikfo: The "logits" are used for both the offline and online version, but the target_logits are only used in the offline and should simply be moved into the if-else logic below.

leakpro/attacks/utils/boosting.py
henrikfo: I'd say they are related, the memorisation just takes the confidence of the model for the correct label while privacy score first rescales the logits using both the correct label and the incorrect ones then normalizes them with the std. One big difference here is that memorization is bounded by -1 and 1 where privacy score is not bounded.

leakpro/attacks/utils/boosting.py
henrikfo: Good question, both works but "or" has been used until now atleast for being easier to code around. But when you imagine which point should be selected, "and" makes more sense. Will change it

leakpro/attacks/utils/boosting.py
henrikfo: This is an expected behaviour for both the online and offline case. For the online case this will happen when you models are not trained enough. The offline model might correctly guess the class with higher confidence than the in model, but this phenomenon will dissapate dissipate with more trained models. In the offline case you will have OUT-samples for the target model that have lower confidences than the average of shadow models.

leakpro/attacks/utils/boosting.py
henrikfo: I think we have to reason around and argue what metric is more important if we should do this. The same reason for why I set the increment to 1.001 for both, setting a higher increment for privacy score showed better performance for the tests I did, but I feel like I dont wanna argue for that disparity

leakpro/attacks/utils/boosting.py
johanos1: But this puts me in an infinite loop because there are not enough samples to satisfy the condition

leakpro/attacks/utils/boosting.py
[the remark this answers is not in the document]
johanos1: yea, that's what I meant

leakpro/attacks/utils/boosting.py
henrikfo: I assume you're building tests for Memorization, how do you initialize the logits/shadow models there?

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+
+        if self.online:
+            for i, (logit, mask, label) in tqdm(enumerate(zip(logits, self.in_indices_mask, self.audit_data_labels)),
+                                                total=len(logits),
+                                                desc="Calculating memorization score",
+                                                leave=False
+                                                ):
+                self.memorization_score[i] = np.mean(logit[mask, label]) - np.mean(logit[~mask, label])
+        else:
```
henrikfo, line 133: What do you think, does this description of the offline implementation make sense? @johanos1 
\# Offline impementation details.
            \# 1. Assumption made: The target logits will be a good approximate to the missing IN-samples
            \# 2. Given that memorization scores of the target model will likely be higher for IN-samples 
            \#       and low for OUT-samples, the offline version does more directly impact the 
            \#       filtering of IN- vs. OUT-samples.

leakpro/attacks/utils/boosting.py
```diff
@@ -0,0 +1,212 @@
...
+                                                ):
+            in_mean, out_mean = np.mean(logit[mask]), np.mean(logit[~mask])
+
+            if len(self.shadow_models) >= 64:
+                in_std, out_std = np.std(logit[mask]), np.std(logit[~mask])
+
+            if self.online:
+                privacy_score.append(np.abs(in_mean-out_mean)/(in_std+out_std+1e-30))
+            else:
+                privacy_score.append(np.abs(target_logit-out_mean)/(2*out_std+1e-30))
```
henrikfo, line 172: And for the privacy score assumptions @johanos1 
\#Assumptions
\# 1. target_logit is a good aproximation for the missing IN-samples
\# 2. out_std ~= in_std => in_std+out_std ~= 2*out_std

johanos1 (approved):

PR #134, opened 2024-09-06
Tabular example

### LINKED ISSUE #95: Generalize QMIA
# Issue
## Problem Description
The quantile regression model is tailored to CIFAR10. This should be general.

## Expected Behavior
The attack should run on arbitrary data.

## What Needs to be Done
Update the assignment of quantile regressor to fit the data type under consideration.

Run QMIA.

### LINKED ISSUE #133: Example for tabular attacks - MIA
# Feature
create example to show how to do attack ona simple tabular dataset

## What Needs to Be Done
Implement the example

## How Can It Be Tested
Run leakpro on top of the dataset

### LINKED ISSUE #135: Make attacks work for models with a single output.
# Issue
## Problem Description
The MIA attacks will crash if the model output is a scalar.

## Expected Behavior
Attacks should run

## What Needs to be Done
Update signals

- Attacks run on example with a single output
- unit test

### LINKED ISSUE #136: Common logger
# Feature
There should be a logger available everywhere in Leakpro

## What Needs to Be Done
Implement common logger

## How Can It Be Tested
Run attacks and make sure logger works.

### LINKED ISSUE #137: Loss trajectory always using label-only
# Issue
## Problem Description
The loss trajectory sets label_only as a string, not a bool. This makes label only always being used.

## Expected Behavior
Use correct setup.

## What Needs to be Done
Change to bool.

Attack should run and correct information used.

### LINKED ISSUE #138: Clean config file
# Feature
The config file is overcrowded with paths, we should not need this many.

## What Needs to Be Done
Remove paths and make them internal

## How Can It Be Tested
ensure attacks run

### LINKED ISSUE #139: Modelhandler blueprint is messy
# Feature
The way we use either target or the provided paths to create a model blueprint is messy.

## What Needs to Be Done
Update and streamline between all model handlers.

## How Can It Be Tested
Attacks should run

johanos1:
# Description

Summary of changes

* Implemented MIA attack example on adult.csv
* Made wrapper for Leakpro to only expose relevant part for user (see leakpro.py)
* Cleaned up the config.yaml (see example)
* Exposed only Leakpro class and AbstractInputHandler to user via __init__.py
* Updated softmax and signals to work for single neuron output.
* Updated loss trajectory to work with user-provided loss functions
* Updated QMIA to use a regression model based on the target model (now it works for all data modalities)
* Created a common logger for the entire package
* Removed folders in config file
* Updated ModelHandler to include replicated code from its children
* Fixed a bug in loss trajectory where label-only was always used.
* Updated tests parameters to always be created newly when a new test is run.
* Included a delete function in ShadowModelHandler for testing purposes

## Resolved Issues

- fixes #133 
- fixes #138 
- fixes #137 
- fixes #136 
- fixes #135 
- fixes #95 
- fixes #139 

## How Has This Been Tested?
Example running.
Unit tests still running.


leakpro/attacks/mia_attacks/abstract_mia.py
```diff
@@ -57,7 +58,6 @@ def __init__(
             AbstractMIA._initialized = True
 
         # These objects are instance specific
-        self.logger = handler.logger
```
henrikfo, line 60: Is there no logger for AbstractMIA now? or is it inherited so its not needed to import?

johanos1: Everywhere where we use the logger, it is imported from the common logger in utils/logger.py. In this particular file, the logger was imported to be used in classes that inherits from abstractMIA. Now, they instead import it directly.

examples/mia/tabular_mia/adult_handler.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__(configs = configs)
+
+
+    def get_criterion(self)->None:
+        """Set the CrossEntropyLoss for the model."""
+        return BCEWithLogitsLoss()
+
+    def get_optimizer(self, model:torch.nn.Module) -> None:
+        """Set the optimizer for the model."""
+        learning_rate = 0.1
```
fazelehh, line 24: Do we want the train hyper parameters to be hard coded for [dataset]_handlers?

examples/mia/tabular_mia/main.ipynb
```diff
@@ -0,0 +1,533 @@
...
+      "text/plain": [
+       "<Figure size 640x480 with 0 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from adult_handler import AdultInputHandler\n",
```
fazelehh, line 490: I get the point of having these imports here, but I still think it is better to move them up and instead write more in the markdown above this cell.

leakpro/attacks/mia_attacks/loss_trajectory.py
fazelehh, line 1: nn.Sequential() is hard coded. We talked about doing something like qmia here.

fazelehh (approved):

PR #150, opened 2024-09-13
Handler rework

### LINKED ISSUE #149: Update the handler creation according to composition pattern
# Feature
We want the user input to be as easy as possible. To this end, we should be able to dynamically create the handler object. This would enable us to:

1. Condition the handler functionality on the user input
2. Extend the handler based on the modality
3. Limit the user input required

## What Needs to Be Done
- Write modality extensions to the input handler
- Dynamically build the handler

## How Can It Be Tested
- Unit tests
- Example running

johanos1:
# Description

Summary of changes

* New and much more flexible design pattern for input handler.
   * Handler now consists of 1) a user-provided input handler, 2) a modality extension

Why is this needed?
A: We will have to tailor certain attacks towards the modality and, therefore, it will be of great aid to dnamically be able to build the handler via composition. This will also allow us to hide, e.g., the metamodule from the user for FL attacks as we can now run different training wrappers depending on FL or MIA.
 
The changes made in this PR are:
* The Leakpro main class now creates the ExtendedHandler that inherits both from the user inputhandler but also from a modality extension class. In this PR, the tabular extension is provided.
* The inputhandler is built dynamically by attaching functionality.
* We now require any dataobject that we read to contain a subset method. This is necessary as we must be able to create replicas of the same data object.
* TabularExtension class is implemented. The assumptions are the following:
   * The dataobject returns either tensor or numpy arrays. Note: we do not allow pandas inputs as it will become a huge pain as we need to use collate functions to convert to tensors in each batch call which is painfully slow. (I tried everything to make this work)...
   * The dataobject contains two alternative components:
      1) a mapping between categorical and one-hot encoded versions of the data (maps the columns)
      2) a flag indicating if the data is one-hot encoded originally or not
* collate functions are added to all dataloaders in case the user wants to provide collate functionality in the data object. If there is none defined, None will be used, i.e., no collate_fn is used.
* Tests are updated accordingly.

## Resolved Issues

- [ ] fixes #149 

## How Has This Been Tested?
- The tabular example on adult is running
- Unit tests implemented and are passing


fazelehh (approved):
the folder name 'user_inputs' is misleading. For me, it sounds like a folder that a user should provide. Maybe something like user_input_handler or user_input_manager is more accurate.

leakpro/leakpro.py
```diff
@@ -1,22 +1,52 @@
...
+                       "graph":None}
 
 class LeakPro:
     """Main class for LeakPro."""
 
-    def __init__(self:Self, handler:AbstractInputHandler, configs_path:str) -> None:
+    def __init__(self:Self, handler_class:AbstractInputHandler, configs_path:str) -> None:
 
-        assert issubclass(handler, AbstractInputHandler), "handler must be an instance of AbstractInputHandler"
+        assert issubclass(handler_class, AbstractInputHandler), "handler must be an instance of AbstractInputHandler"
```
henrikfo, line 49: Really like this way of coding, more checks in runtime and you can get them for free in testing aswell. Maybe worth considering to extend this practise?

henrikfo (approved):

examples/mia/tabular_mia/utils/adult_data_preparation.py
```diff
@@ -74,27 +79,45 @@ def get_adult_dataset(path):
...
+        for i, feature in enumerate(numerical_features):
+            dec_to_onehot_mapping[i] = [x_final.columns.get_loc(feature)]  # Mapping to column index
+
+        # Add one-hot encoded features to the dictionary
+        for i, categorical_feature in enumerate(categorical_features):
+            j = i + len(numerical_features)
+            one_hot_columns = [col for col in one_hot_feature_names if col.startswith(categorical_feature)]
+            dec_to_onehot_mapping[j] = [x_final.columns.get_loc(col) for col in one_hot_columns]
+
+
```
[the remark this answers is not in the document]
johanos1: What do you mean not a common format? Is there a common format to map between categorical and one-hot encoded?

leakpro/leakpro.py
johanos1: I think we follow this quite a bit throughout but we do not commonly use asserts but exceptions.

PR #159, opened 2024-10-11
Working starting point for FL

viktorvaladi:
Rough starting point that needs some more work for FL. The attack is working here and moved to example/ with some refactoring / additions on functionality


examples/gia/cifar100_inverting_16_images/model.py
mattiasakesson, line 1: can't we have models in a subdirectory under examples to get rid of duplications?

examples/gia/cifar10_inverting_1_image/cifar.py
mattiasakesson, line 1: can't we have dataset in a subdirectory under examples to get rid of duplications for every example?

leakpro/attacks/gia_attacks/invertinggradients.py
mattiasakesson, line 1: maybe add the individual settings in init for class InvertingGradients?

mattiasakesson (changes requested):
I added some design questions, but overall everything looks good.

leakpro/attacks/gia_attacks/invertinggradients.py
viktorvaladi: Could you elaborate?

examples/gia/cifar100_inverting_16_images/model.py
viktorvaladi: A good point, but I would argue that for transparency it's better to have each example completely standalone with as few dependencies to our framework as possible. So each example folder should clearly present the parts that we expect a user to supply to run LeakPro. I think in this case repetition of code is OK

examples/gia/cifar10_inverting_1_image/cifar.py
viktorvaladi: Same as above (and also the cifars vary a bit in their implementation)

leakpro/attacks/gia_attacks/invertinggradients.py
mattiasakesson: like 
InvertingConfig(total_variation = 1.0e-06, attack_lr = 0.1, at_iterations = 8000, optimizer = MetaSGD(),criterion = CrossEntropyLoss(), epochs = 1):
   self.total_variation=total_variation
   ...
then you could have the possibility to build the object in one line:
InvertingConfig(kvarg)

viktorvaladi: Good point. I pushed adding types to the config. Then the @dataclass decorator handles the rest, so now we can initialize it like this:
`config = InvertingConfig(total_variation = 1.0e-04)`
:)

mattiasakesson (approved):

PR #161, opened 2024-10-14
Fix micromamba command

fazelehh:
# Description

Issue: The GitHub Actions workflow was failing due to an unexpected -p argument being passed to the micromamba command.

Changes:
Added a step to install micromamba using a curl command.
Updated the micromamba command to create the environment without the -p argument.

## How Has This Been Tested?
Verified that the workflow runs successfully with the changes.

## Related Pull Requests

- #PR


viktorvaladi (approved):
Gj!

PR #208, opened 2025-01-17
Optuna

viktorvaladi:
Adding new generic setup for attack objects to use optuna.

An attack object needs three functions to use this method:

reset_attack() # resets an attack to its initial state
suggest_parameters(trial) # gives a range and specifies which parameters to optimize
get_configs() # get attack config

Also this PR abstract out some functionality from GIA's to reduce code duplication.

### DIFF
--- a/examples/gia/huang_optuna/main.py
+++ b/examples/gia/huang_optuna/main.py
@@ -2,16 +2,16 @@
 
 from cifar import get_cifar10_loader
 
+from leakpro.attacks.gia_attacks.huang import Huang, HuangConfig
 from leakpro.fl_utils.gia_train import train
-from leakpro.run import huang_optuna
+from leakpro.optuna import optuna_optimal_hyperparameters
 from model import ResNet, PreActBlock
 
 if __name__ == "__main__":
-    # Instantiate the base model and get the CIFAR-10 loader
-
-    # Pre activation required for the attack to give decent results
+    # Pre activation required for this attack to give decent results
     base_model = ResNet(PreActBlock, [2, 2, 2, 2], num_classes=10)
     cifar10_loader, mean, std = get_cifar10_loader(num_images=16, batch_size=16, num_workers=2)
 
     # Run Optuna optimization with Huang
-    huang_optuna(base_model, cifar10_loader, train, mean, std)
+    attack_object = Huang(base_model, cifar10_loader, train, mean, std, HuangConfig())
+    optuna_optimal_hyperparameters(attack_object=attack_object)
--- a/examples/gia/huang_optuna/main.py
+++ b/examples/gia/huang_optuna/main.py
@@ -4,7 +4,7 @@ from cifar import get_cifar10_loader
 
 from leakpro.attacks.gia_attacks.huang import Huang, HuangConfig
 from leakpro.fl_utils.gia_train import train
-from leakpro.optuna import optuna_optimal_hyperparameters
+from leakpro.hyperparameter_tuning.optuna import optuna_optimal_hyperparameters
 from model import ResNet, PreActBlock
 
 if __name__ == "__main__":
--- a/examples/gia/huang_optuna/main.py
+++ b/examples/gia/huang_optuna/main.py
@@ -2,16 +2,14 @@
 
 from cifar import get_cifar10_loader
 
-from leakpro.attacks.gia_attacks.huang import Huang, HuangConfig
-from leakpro.fl_utils.gia_train import train
-from leakpro.hyperparameter_tuning.optuna import optuna_optimal_hyperparameters
+from leakpro.attacks.gia_attacks.huang import Huang
 from model import ResNet, PreActBlock
 
 if __name__ == "__main__":
     # Pre activation required for this attack to give decent results
     base_model = ResNet(PreActBlock, [2, 2, 2, 2], num_classes=10)
-    cifar10_loader, mean, std = get_cifar10_loader(num_images=16, batch_size=16, num_workers=2)
+    client_loader, mean, std = get_cifar10_loader(num_images=16, batch_size=16, num_workers=2)
 
     # Run Optuna optimization with Huang
-    attack_object = Huang(base_model, cifar10_loader, train, mean, std, HuangConfig())
-    optuna_optimal_hyperparameters(attack_object=attack_object)
+    attack_object = Huang(base_model, client_loader, mean, std)
+    attack_object.run_with_optuna()
--- a/leakpro/attacks/gia_attacks/abstract_gia.py
+++ b/leakpro/attacks/gia_attacks/abstract_gia.py
@@ -1,6 +1,6 @@
 """Module that contains the abstract class for constructing and performing a membership inference attack on a target."""
 
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from collections.abc import Generator
 from copy import deepcopy
 from typing import Callable, Optional
--- a/leakpro/attacks/gia_attacks/abstract_gia.py
+++ b/leakpro/attacks/gia_attacks/abstract_gia.py
@@ -69,7 +69,8 @@ class AbstractGIA(ABC):
 
         Returns
         -------
-            Generator with intermediary results and final GiaResults.
+            Generator with intermediary results to allow for lazy evaluation during
+            hyperparameter turning, and final GiaResults.
 
         """
         pass
@@ -95,6 +96,7 @@ class AbstractGIA(ABC):
                             ) -> Generator[tuple[int, Tensor, GIAResults]]:
         """Generic attack loop for GIA's."""
         optimizer = torch.optim.Adam([reconstruction], lr=attack_lr)
+        # reduce LR every 1/3 of total iterations
         scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer,
                                                             milestones=[at_iterations // 2.667,
                                                                         at_iterations // 1.6,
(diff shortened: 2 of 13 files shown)


leakpro/optuna.py
```diff
@@ -0,0 +1,56 @@
...
+                # save results if not pruned
+                if result_object is not None:
+                    result_object.save(name="optuna", path="./leakpro_output/results", config=attack_object.get_configs())
+                    return intermediary_results
+        elif isinstance(result, MIAResult):
+            return result.accuracy # add something reasonable to optimize toward here
+        return None
+
+    # Define the pruner and study
+    pruner = optuna.pruners.MedianPruner(n_warmup_steps=n_warmup_steps)
```
johanos1, line 39: allow for different pruners and samplers on the input?

leakpro/optuna.py
```diff
@@ -0,0 +1,56 @@
...
+from torch import Tensor
+
+from leakpro.attacks.gia_attacks.abstract_gia import AbstractGIA
+from leakpro.attacks.mia_attacks.abstract_mia import AbstractMIA
+from leakpro.metrics.attack_result import MIAResult
+from leakpro.utils.logger import logger
+from leakpro.utils.seed import seed_everything
+
+
+def optuna_optimal_hyperparameters(attack_object: Union[AbstractGIA, AbstractMIA], n_warmup_steps:int= 5,
```
johanos1, line 15: Make an assess to ensure attack_object is the correct type.

pyproject.toml
```diff
@@ -20,6 +20,9 @@ dependencies = [
   "joblib",
   "pyyaml",
   "scikit-learn",
+  "pytorch-ignite",
```
johanos1, line 23: ignite should only be in `dev` and `federated`

pyproject.toml
```diff
@@ -20,6 +20,9 @@ dependencies = [
...
+  "gdown",
+  "optuna",
```
johanos1, line 25: optuna should not be in synthetic data as of now

pyproject.toml
```diff
@@ -20,6 +20,9 @@ dependencies = [
   "joblib",
   "pyyaml",
   "scikit-learn",
+  "pytorch-ignite",
+  "gdown",
```
johanos1, line 24: why do we need gdown here? Is it for the examples? If so, it should be specified in there and not be installed in Leakpro.

leakpro/fl_utils/similarity_measurements.py
```diff
@@ -1,13 +1,38 @@
...
 from ignite.metrics import SSIM
-from torch import Tensor, abs, cuda, mean, nn, no_grad, norm
-from torch.nn.modules.utils import _pair, _quadruple
+from torch import Tensor, abs, cuda, mean, no_grad, norm
 from torch.utils.data import DataLoader
 from torchmetrics.functional import peak_signal_noise_ratio
 
-from leakpro.utils.import_helper import Self
 
+def cosine_similarity_weights(client_gradient: torch.Tensor, reconstruction_gradient: torch.Tensor,
```
johanos1, line 9: Nice. So now both inverting and Huang use the same function. Did you test that both still work?

leakpro/attacks/gia_attacks/invertinggradients.py
```diff
@@ -157,42 +126,34 @@ def closure() -> torch.Tensor:
...
-            filtered_input_gradients = [client_gradient[i] for i in indices]
-        costs = sum((x * y).sum() for x, y in zip(filtered_input_gradients,
-                                                  filtered_trial_gradients))
-
-        trial_norm = sum(x.pow(2).sum()
-                         for x in filtered_trial_gradients).sqrt()
-        input_norm = sum(y.pow(2).sum()
-                         for y in filtered_input_gradients).sqrt()
-        return 1 - (costs / trial_norm / input_norm)
+    def reset_attack(self: Self) -> None:
```
johanos1, line 148: Can we make it so this is called in the constructor? It would make sense to reset when creating the object?

leakpro/attacks/gia_attacks/huang.py
```diff
@@ -172,46 +142,43 @@ def closure() -> torch.Tensor:
...
-            filtered_input_gradients = [client_gradient[i] for i in indices]
-        costs = sum((x * y).sum() for x, y in zip(filtered_input_gradients,
-                                                  filtered_trial_gradients))
-
-        trial_norm = sum(x.pow(2).sum()
-                         for x in filtered_trial_gradients).sqrt()
-        input_norm = sum(y.pow(2).sum()
-                         for y in filtered_input_gradients).sqrt()
-        return 1 - (costs / trial_norm / input_norm)
+    def reset_attack(self: Self) -> None:
```
johanos1, line 173: Could we call this also in the constructor

leakpro/attacks/gia_attacks/invertinggradients.py
```diff
@@ -157,42 +126,34 @@ def closure() -> torch.Tensor:
...
-        This function calculates the pairwise costs between each client gradient and the reconstruction gradient
-        using the cosine similarity measure. The costs are accumulated and averaged over all client gradients.
+    def _configure_attack(self: Self, configs: dict) -> None:
+        pass
 
-        Returns
-        -------
-            torch.Tensor: The average reconstruction cost.
+    def suggest_parameters(self: Self, trial: optuna.trial.Trial) -> None:
+        """Suggest parameters to chose and range for optimization for the Huang attack."""
```
johanos1, line 144: Replace "Huang attack" with Inverting Gradient

leakpro/attacks/gia_attacks/huang.py
```diff
@@ -172,46 +142,43 @@ def closure() -> torch.Tensor:
...
             loss_r_feature = sum([
                 mod.r_feature
                 for (idx, mod) in enumerate(self.loss_r_feature_layers)
             ])
 
             # Add the TV loss term to penalize large variations between pixels, encouraging smoother images.
-            rec_loss += self.t_v_scale * total_variation(self.reconstruction)
-            rec_loss += self.bn_reg * loss_r_feature
-            rec_loss += self.l2_scale * l2_norm(self.reconstruction)
+            rec_loss += self.configs.total_variation * total_variation(self.reconstruction)
```
johanos1, line 155: I thought this was a bool. Please rename to something with scale or reg

johanos1 (changes requested):

johanos1:
Please also link the relevant issues in the PR.

examples/gia/huang_optuna/main.py
```diff
@@ -2,16 +2,16 @@
 
 from cifar import get_cifar10_loader
 
+from leakpro.attacks.gia_attacks.huang import Huang, HuangConfig
 from leakpro.fl_utils.gia_train import train
```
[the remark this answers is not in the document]
viktorvaladi: There's no run function here, the optuna method takes an attack object as input

examples/gia/huang_optuna/main.py
```diff
@@ -2,16 +2,16 @@
 
 from cifar import get_cifar10_loader
 
+from leakpro.attacks.gia_attacks.huang import Huang, HuangConfig
```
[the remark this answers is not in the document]
viktorvaladi: I think exposing the object to users is a decent level of abstraction, to allow for things like overriding functionality etc. I'm thinking of moving away from the "run methods" for GIA, to instead have the default being to expose the attack objects. 

But maybe it's okay to have leakpro.run("huang") as a lazy option as well, could investigate it

leakpro/attacks/gia_attacks/abstract_gia.py
```diff
@@ -57,17 +64,66 @@ def prepare_attack(self:Self) -> None:
...
+                    )
+                if (i +1) % 500 == 0 and median_pooling:
+                    reconstruction.data = MedianPool2d(kernel_size=3, stride=1, padding=1, same=False)(reconstruction)
+            # Chose image who has given least loss
+            if loss < self.best_loss:
+                self.best_loss = loss
+                self.best_reconstruction = deepcopy(reconstruction_loader)
+                self.best_reconstruction_round = i
+                logger.info(f"New best loss: {loss} on round: {i}")
+            if i % 250 == 0:
```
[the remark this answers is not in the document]
viktorvaladi: Yes exactly

leakpro/optuna.py
viktorvaladi: I think if we do that then we should allow for different optuna.pruners, optuna.samplers and optuna.storage. But I'm not sure how they work with e.g. hyperparameters etc - i think this can be future work for when we have more examples with optuna and know what we might want to vary?

leakpro/fl_utils/similarity_measurements.py
viktorvaladi: Yep works for both of them. I really don't understand why though. But whatever works works i guess

leakpro/attacks/gia_attacks/invertinggradients.py
viktorvaladi: Would it? It would already be in its initial state so it would just call the same things twice right?

leakpro/attacks/gia_attacks/huang.py
viktorvaladi: as above

examples/gia/huang_optuna/main.py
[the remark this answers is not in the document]
johanos1: Sure, but doesnt it make more sense to just send `base_model, cifar10_loader, mean, std` into Leakpro. That way you can use the same interface with other attacks. Then you can have a flag to decide to use Optuna? Also a string for the attack?

examples/gia/huang_optuna/main.py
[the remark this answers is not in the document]
johanos1: What is the benefit of exposing it like this? I mean, why would you want to override the functionality of Huang?

leakpro/optuna.py
johanos1: fair enough.

leakpro/fl_utils/similarity_measurements.py
johanos1: I guess the sign differ in Geiping and it therefore works with a max..

johanos1 (approved):
I like the update!

PR #227, opened 2025-02-13
add the new arguments

mirmusss:
# Description

Summary of changes

* ... 
* ...

## Resolved Issues

- [ ] fixes #Issue

## Related Pull Requests

- #PR


leakpro/synthetic_data_attacks/anonymeter/evaluators/singling_out_evaluator.py
```diff
@@ -209,103 +232,195 @@ def univariate_singling_out_queries(*, df: pd.DataFrame, n_queries: int) -> List
...
+            continue
+
+        if isinstance(dtypes[col], pd.CategoricalDtype) and is_numeric_dtype(dtypes[col].categories.dtype):
+            query.append(f"{col} == {record[col]}")
         else:
-            item = f'== "{record[col]}"'
-        query.append(f"{col} {item}")
+            if not use_medians:
+                value_counts: pd.Series = df[col].value_counts()
+                if value_counts.get(record[col], 0) == 1:
```
gpadres, line 298: there is no else here so this path is not going to append anything to query if value_counts.get != 1. I dont fully understand how can we end up here (e.g. record[col] has to be non-numerical, non (categorical numerical), so like string categories? maybe we can discuss this

gpadres: also i suppose the not use_medians is for handling the case above of use_medians? i dont fully understand why we need to distinguish about medians here if these are not numerical attributes

leakpro/synthetic_data_attacks/anonymeter/evaluators/singling_out_evaluator.py
```diff
@@ -209,103 +232,195 @@ def univariate_singling_out_queries(*, df: pd.DataFrame, n_queries: int) -> List
...
+                query.append(f"{col} {operator} {record[col]}")
             else:
-                operator = "<="
-            item = f"{operator} {record[col]}"
-        #Categorical
-        elif isinstance(dtypes[col], pd.CategoricalDtype) and is_numeric_dtype(dtypes[col].categories.dtype):
-            item = f"== {record[col]}"
-        elif isinstance(record[col], str):
-            item = f"== '{escape_quotes(input=record[col])}'"
+                values: pd.Series = df[col].dropna().sort_values()
```
gpadres, line 281: sort_values and value_counts are expensive operations so this would need to be handled outside this function (since we need it for all the columns and only once). I like the idea of modifying to consider for medians or not.

leakpro/synthetic_data_attacks/anonymeter/evaluators/singling_out_evaluator.py
```diff
@@ -209,103 +232,195 @@ def univariate_singling_out_queries(*, df: pd.DataFrame, n_queries: int) -> List
...
+                        query.append(f"{col} <= {values.iloc[0]}")
+                    elif record[col] >= values.iloc[-1]:
+                        query.append(f"{col} >= {values.iloc[-1]}")
+                    else:
+                        value_counts: pd.Series = df[col].value_counts()
+                        if value_counts.get(record[col], 0) == 1:
+                            query.append(f"{col} == {record[col]}")
+            continue
+
+        if isinstance(dtypes[col], pd.CategoricalDtype) and is_numeric_dtype(dtypes[col].categories.dtype):
```
gpadres, line 293: i see now that we can improve the original code, i thought we were taking into consideration numerical categorical values but i see now that the way the code is right now we are not. lets dicuss

PR #242, opened 2025-02-25
added inheritance for shadow models

### LINKED ISSUE #158: Consistent loss and acc difference between shadow models and target model
# Issue
## Problem Description
Consistent loss and acc difference between shadow models and target model during training.

## Expected Behavior
For many runs of leakpro, training taget and shadow model from scratch, they should have the same average of loss and acc.
Right now the target model is on average 2-3 % points better than the average of shadow models trained.

## What Needs to be Done
Figure out the difference within the training loop, amount of data, what data the target model is trained on etc.

Run leakpro with retrain=True and train and some shadow models for 6 or more epochs a few times

henrikfo:
# Description

Summary of changes
Added inheritance of target model config for shadow models. Lets the shadow models inherit optimizer and criterion from the target model.

## Resolved Issues
- Tries to resolve #158
        Not fully successful.
 
## How Has This Been Tested?
A few different optimizers have been successfully inherited.


henrikfo:
This PR and implementation of inheritance has a similar solution in and is pushed with PR #269

PR #262, opened 2025-03-14
Singling out

mirmusss:
# Detailed Analysis of Key Improvements to the Singling-Out Attack Algorithm

This update refines our multivariate singling-out attack algorithm by introducing several key improvements that enhance query diversity, computational efficiency, and adaptability to different data characteristics. In particular, we now incorporate a decision tree–based query construction for continuous columns to generate adaptive, data-driven intervals.

---

### Query Generation Strategy

- **Original Approach:**  
  The original version relied on simple random sampling for records and columns. This approach was prone to missing important patterns in the data.

- **Improved Approach:**  
  - Uses `itertools.combinations` to systematically explore all possible column combinations, ensuring better coverage and diversity of queries.
  - Integrates both naive and decision tree–based methods to construct predicates that isolate individual records more effectively.

### Progress Tracking and Efficiency

- **New Parameters Introduced:**
  - **`max_per_combo`**: Maximum queries allowed per column combination.
  - **`sample_size_per_combo`**: Number of records sampled from each combination.
  - **`max_rounds_no_progress`**: Number of rounds without finding new queries before early stopping.
  - **`max_attempts`**: Total number of attempts allowed to generate queries.

- **Benefits:**
  - Better management of computational resources.
  - Clear progress tracking and early stopping to prevent infinite loops.
  - Fine-grained control over query generation to adapt to various data properties.

---

#### Previous Version:
```python
def query_from_record(*,
    record: pd.Series,
    dtypes: pd.Series,
    columns: List[str],
    medians: Optional[pd.Series]
) -> str:

# Improved `query_from_record` Function

## Improved Version

```python
def query_from_record(*,
    df: pd.DataFrame,
    record: pd.Series,
    dtypes: pd.Series,
    columns: List[str],
    medians: Optional[pd.Series] = None,
    use_medians: bool = True,
    precomputed_sorted: Optional[dict[str, pd.Series]] = None,
    precomputed_counts: Optional[dict[str, pd.Series]] = None,
    use_tree: bool = False,
    tree_models: Optional[Dict[str, any]] = None
) -> str:
    # Implementation goes here...
    # When use_tree is True and a tree model exists for a continuous column,
    # the decision tree’s decision path is used to derive an adaptive interval.
    # Otherwise, the function falls back on median-based or value-based logic.
```
# Key Enhancements

1. **use_medians Toggle**  
   **Description:**  
   Allows switching between median-based conditions and value-based (extreme value/uniqueness) checks. This provides flexibility in how queries are constructed for numeric attributes.

2. **Direct DataFrame Access**  
   **Description:**  
   The function now accepts the full DataFrame (`df`) to build queries based on actual data values. This ensures accurate comparisons when constructing predicates.

3. **Improved Numeric and Categorical Handling**  
   **Description:**  
   Enhances robustness by ensuring that numeric and categorical data types are handled appropriately:
   - For numeric columns, it supports both median-based and extreme value/uniqueness checks.
   - For categorical data, it includes improved uniqueness checking to help isolate individual records.

4. **Integration with Decision Trees**  
   **When `use_tree=True`:**  
   For continuous columns, a pre-fitted decision tree (stored in a dictionary keyed by column name) is used to derive an adaptive interval.
   
   - **Pre-Fitting:**  
     For each continuous column (excluding booleans), rows with missing values are dropped and a univariate `DecisionTreeRegressor` is fitted using the column as both feature and target.
   
   - **Decision Path Extraction:**  
     For a given record, the decision tree’s decision path is used to extract the thresholds along the path. These thresholds define the lower and upper bounds of an interval that ideally isolates the record.
   
   - **Adaptive Interval Query:**  
     The derived interval is formatted as a query (e.g., `"col >= lower_bound & col < upper_bound"`), providing a data-driven predicate that adapts to the underlying data distribution.


mirmusss:
https://github.com/aidotse/LeakPro/blob/81a56e947fdf43a7c5abe50b8dede5841e8c9af5/leakpro/synthetic_data_attacks/anonymeter/evaluators/singling_out_evaluator.py#L407 

this moves the sorting outside of the query from the record so that this operation is only done once

mirmusss:
> there is no else here so this path is not going to append anything to query if value_counts.get != 1. I dont fully understand how can we end up here (e.g. record[col] has to be non-numerical, non (categorical numerical), so like string categories? maybe we can discuss this

The absence of an else branch in that path is intentional. The design is that if the value isn’t unique (i.e. if value_counts.get(record[col], 0) != 1), no condition is added for that attribute. In other words, if the record’s value isn’t at one of the extremes (min or max) and isn’t unique, it isn’t useful for singling out that record, so nothing is appended to the query. This behavior is consistent with the univariate attack logic, where only conditions that truly single out a record (e.g., a unique value) are considered.    This logic is essentially the univariate singling‐out approach shown in pseudo‐code snippet (“Algorithm 1”) refer to the paper. The univariate approach focuses on:
* Checking NaNs (singling out if exactly one record is NaN),
* Using min/max for numeric attributes (i.e., a <= min(...) and a >= max(...)),
* Checking for unique (rare) values in the column, appending a == v only if v appears once.
And just for the clarity in Python, that translates to:
1. df[col].isna().sum() == 1 → “a Is NaN” predicate.
2. col <= min_val or col >= max_val → extreme numeric predicates.
3. value_counts == 1 → “a == v” if it’s a unique value.
So, the “no‐else” code path is effectively capturing “Algorithm 1”’s line “if Sum(X_syn[:, a] == v) == 1 then.” If the sum is not 1, we do nothing—exactly matching the univariate logic that only rare (unique) values matter for singling out.

henrikfo:
Closing this PR when it is included in #364

PR #268, opened 2025-03-17
Fix numerical categorical cols singling out

mirmusss:
# Pull Request Description

## Description
This pull request updates the behavior of the `SinglingOutEvaluator` and its associated query generation functions to correctly handle low-cardinality and binary columns. The changes ensure that:
- Numeric columns with 20 or fewer unique values are converted to categorical types.
- Binary numeric columns (with values exactly `{0, 1}`) are converted to boolean types.
- The query generator (`query_from_record`) now uses equality (`==`) for categorical (or boolean) columns and numeric operators (such as `>=` or `<=`) for numeric columns with more than 20 unique values.
- Test functions are updated accordingly so that multivariate queries are generated based on the processed data types.
- The storage directory teardown in the test fixture is updated to remove non-empty directories.

## Summary of Changes
- **Evaluator Updates:**  
  - Modified `SinglingOutEvaluator.__init__` to:
    - Drop duplicate rows.
    - Convert numeric columns with low unique counts (<= 20) to categorical.
    - Convert binary columns (values `{0, 1}`) to boolean.
- **Query Generation Adjustments:**  
  - Updated `query_from_record` so that:
    - Columns with categorical types (or booleans) generate equality comparisons.
    - Numeric columns (with more than 20 unique values) generate comparisons using randomly chosen operators (`>=` or `<=`).
- **Test Updates:**  
  - Updated the `test_multivariate_singling_out_queries` to instantiate the evaluator first, ensuring the proper type conversions are applied.
  - Added tests using regex patterns to validate that (the reason for this instead of the simple changing of the values for column c0 in the old version is that, the c1 was being considered as categorical too):
    - For categorical columns, queries follow the pattern: `c0 == <integer>` (with optional `.0` formatting).
    - For numeric columns, queries follow the pattern: `c1 >=|<= <number>` (with optional decimal part).
## Resolved Issues
- Fixes #Issue  

## How Has This Been Tested?
- Ran the complete test suite using:
  ```bash
  pytest --maxfail=1 --disable-warnings -q


gpadres (approved):

henrikfo (changes requested):
When we resolve the comments we can merge!

gpadres:
Analysis on results:
IBM HR:
main attack seems more smooth (and higher values)
naive attack underperforming from before
-> higher excess risk

adults:
main attack not as smooth 
naive attack slightly underperforming
-> slightly higher excess risk

texas:
main attack smoother, and way higher values
naive attack way lower values
although a lot of columns excess risk is corrected (>0) 
there are still a couple of remaining columns with excess risk <=0 (3,4,5)

census:
same pattern, main attack smoother (and higher values)
naive attack underperfroming from before
-> higher excess risk


Getting issue with warning when setting category

Action points:
1. warning when setting category, see if we can take away (if possible identify on which cases)
2. review naive attack code, see if maybe we need to change the order or similar of the construction of the query .
(otherwise why is naive attack performing worse?)

further work:
why is texas excess risk <=0 for some columns?

Issue #275, opened 2025-03-18
TPR@0.0%FPR is always N/A in MiaResult


totte174:
# Issue
TPR@0.0%FPR is always N/A in MiaResult

## Problem Description
TPR@0.0%FPR is always N/A in MiaResult the current functionality check for FPR below threshold. Since fpr < 0.0 is impossible it always returns N/A.

## Expected Behavior
It should check if less than or equal to threshold.

PR #276, opened 2025-03-18
Fix tpr@0.0%fpr to not be N/A

totte174:
# Description
TPR@0.0%FPR is always N/A in MiaResult the current functionality check for FPR below threshold. Since fpr < 0.0 is impossible it always returns N/A.

## Resolved Issues
#275

## Related Pull Requests

- #PR


leakpro/metrics/attack_result.py
```diff
@@ -718,8 +718,8 @@ def get_result_fixed_fpr(fpr: list, tpr: list) -> dict:
     def find_tpr_at_fpr(fpr_array:np.ndarray, tpr_array:np.ndarray, threshold:float) -> float:
         """Find tpr for a given fpr."""
         try:
-            # Find the last index where FPR is less than the threshold
-            valid_index = np.where(fpr_array < threshold)[0][-1]
+            # Find the last index where FPR is less than or equal to the threshold
+            valid_index = np.where(fpr_array <= threshold)[0][-1]
```
fazelehh, line 722: can you also add a if condition to check if valid_index is empty or not, and if yes, handle it and return nan .
Also, it would be great if you also fix the docstring in line 716. It is """Find TPR values for fixed TPRs.""", it should be """Find TPR values for fixed FPR."""

fazelehh (changes requested):

leakpro/metrics/attack_result.py
johanos1: Can this ever be empty of threshold and fpr_array are not? I guess it depends on how threshold is cretaed.

totte174: I made some changes to address this and some other issues I had.

1. Fixed handling where no fpr <= threshold is found. This was already handled before but now returns nan instead of a string.
2. When creating fpr and tpr array, dont include entries where denominator is zero. Before these threshold would result in zero-filled entry which also would mess up when trying to find tpr@0fpr.
3. Sort FPR array as done in CombinedMetricResult.

I have not tested this yet and will report back when I have ran an attack.

leakpro/metrics/attack_result.py
```diff
@@ -711,19 +720,17 @@ def _latex(results: list) -> str:
...
-            valid_index = np.where(fpr_array < threshold)[0][-1]
-            return float(f"{tpr_array[valid_index] * 100:.4f}")
-        except IndexError:
-            # Return None or some default value if no valid index found
-            return "N/A"
+    
+def find_tpr_at_fpr(fpr_array: np.ndarray, tpr_array:np.ndarray, threshold:float) -> float:
+    """Find TPR for a given FPR."""
+    # Find the last index where FPR is less than or equal to the threshold
+    valid_index = np.nonzero(fpr_array <= threshold)[0]
```
henrikfo, line 727: Is there a reason for the change from np.where to np.nonzero?

leakpro/metrics/attack_result.py
```diff
@@ -711,19 +720,17 @@ def _latex(results: list) -> str:
...
+    if len(valid_index) == 0:
+        return np.nan
+    return float(f"{tpr_array[valid_index] * 100:.4f}")
```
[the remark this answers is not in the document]
totte174: Yes sorry, again I haven't tested this yet.

leakpro/metrics/attack_result.py
henrikfo: Ok, sounds good!

leakpro/metrics/attack_result.py
```diff
@@ -711,19 +720,17 @@ def _latex(results: list) -> str:
...
-        except IndexError:
-            # Return None or some default value if no valid index found
-            return "N/A"
+    
+def find_tpr_at_fpr(fpr_array: np.ndarray, tpr_array:np.ndarray, threshold:float) -> float:
+    """Find TPR for a given FPR."""
+    # Find the last index where FPR is less than or equal to the threshold
+    valid_index = np.nonzero(fpr_array <= threshold)[0]
+    if len(valid_index) == 0:
+        return np.nan
```
henrikfo, line 729: Also, would it makes more sense to return a 0.0 instead of np.nan / "N/A"? 
The logic being that if no there is no FPR less-than-or-equal-to e.g. 0.01%, then there will be no TPR values to plot further left (for lower FPRs).
If that case, the corresponding ROC curve would extend from the last TPR value at the last FPR value, horizontally through the plot to the left. Falsely implying a stronger attack.
Setting a return value of 0.0 instead would instead let ROC curve drop straight down (to TPR=0.0)

leakpro/metrics/attack_result.py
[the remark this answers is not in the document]
totte174: I have fixed this and tested running an MI-attack, it seems to work well.

henrikfo (approved):

johanos1:
@henrikfo do you mind fixing the tests for this PR?

henrikfo:
For sure!

henrikfo (dismissed):

henrikfo (approved):

fazelehh (approved):

PR #289, opened 2025-03-23
Reporting Update

### LINKED ISSUE #282: Refactor MIAResult and FPR/TPR calculation
# Feature
Refactor MIAResult and FPR/TPR calculation

Right now the creating predictions and using them to calculate FPR / TPR is handled in part by the attacks themselves (creating a linspace of threshold values ect..). This code is generally reused for all attacks (except for checking if score is greater or less than a threshold) and should be handled within MIAResult since it is not specific to the attacks.
Secondly the way we calculate FPR/TPR right now is very unstable can fail completely if outlier membership scores are present (due to creating a linspace between min and max MIAScore).

Example of signal histogram with MIAScore outliers:
Image

Instead I propose that an attack should provide to MIAResult the following:

- An array of audit indices sorted by MIAScore where a lower position in array implies higher predicted membership.
- The true labels of the audit points.
- MIAScores used for creating signal histogram.
- Possible probabilites of prediction for attacks which supply this (predictions_proba).

Instead of just looking at 1000 possible thresholds which we currently do for most attack, we should go through the sorted audit indices sequentially and calculate the fpr and tpr for every single MIAScore/signal value.
I believe that this will ensure that the ROC curve plot and tpr_at_fixed_fpr table will always be accurate even for outliers in attacks, something which is very important during auditing and development of attacks.

## What Needs to Be Done

1. Refactor MIAResult class to take in the parameters described above at initialization and calculate fpr/tpr by sequentially going through the sorted indice list. (Be very careful here that calculation is correct)
2. Refactor current membership inference attacks to sort indices by MIAScore and and supply this to MIAResult.
3. Possibly look at taking a subset of fpr/tpr values when creating data.json file. Using all possible MIAScore values as thresholds can result in a 1000x larger data.json file (if for example using a dataset with 1M points).

## How Can It Be Tested
Run current examples with the updated attacks and see if resulting ROC curve and tpr@fixed fpr values are the same (or very similar).

### LINKED ISSUE #287: Saving MIAResult to disk crashes when return_results=False
# Issue
Saving MIAResult to disk crashes when return_results=False

## Problem Description
It gives the following error indicating this function is not modified for changes to AuditConfig

`TypeError
[9] leakpro = LeakPro(CifarInputHandler, config_path)
[11] # Run the audit 
---> [12] mia_results_optuna = leakpro.run_audit(return_results=False)

File LeakPro\leakpro\leakpro.py:101, in LeakPro.run_audit(self, return_results, use_optuna)
[99]  else:
[100] result = audit_results[attack_name]["result_object"]
--> [101]         result.save(name=attack_name, path=self.report_dir, config=self.handler.configs.audit)
[103] logger.info("Auditing completed")
[104] return results

File LeakPro\leakpro\metrics\attack_result.py:274, in MIAResult.save(self, path, name, config, show_plot)
[271] def save(self:Self, path: str, name: str, config:dict = None, show_plot:bool = False) -> None:
[272]     """Save the MIAResults to disk."""
--> [274]     result_config = config["attack_list"][name]
[275]     fixed_fpr_table = get_result_fixed_fpr(self.fpr, self.tpr)
[277]     # Get the name for the attack configuration

TypeError: 'AuditConfig' object is not subscriptable`

## Expected Behavior
Save results correctly.

## What Needs to be Done
Update function for current AuditConfig implementation

Run any attack with run_audit(return_results=False) and check that result are correctly saved.

### LINKED ISSUE #288: Dead code in report handler
# Issue
## Problem Description

There is alot of legacy code int he report handler.

## Expected Behavior
The code should be removed.

## What Needs to be Done
Clean the repo from unused code.

### LINKED ISSUE #290: Docstrings are no longer enforced by ruff
# Issue
## Problem Description
The checks for docstrings seems to be no longer supported

## Expected Behavior
Docstrings should be enforced

## What Needs to be Done
Add an explicit check for this, potentially using pydocstyle

Ensure the docstrings are enforced.

### LINKED ISSUE #291: Use Schemas in Result classes
# Issue
## Problem Description
Right now, it is hard to follow what should go into a result object.

## Expected Behavior
It should follow a predefined schema

## What Needs to be Done
Incorporate schemas with the result classes

Tests pass.

johanos1:
# Description

In #287 , #282 , #288, there have been several things pointed out regarding the report handler. We should update it to fix these.

Summary of changes

* Cleaned all components in report handler that are not in use
* Updated the thresholding for ROC computation to be more granular
* Froze ruff version to 0.11.2
   * Note: the updated ruff has removed AN101, i.e., enforcing Self and cls types.
   * It does **not** enforce that docstrings are included. It does, however, ensure that docstrings follow the correct format once included.
* Moved optuna objective for MIA to be set in the AbstractMIA. This was to enable the use of schemas in MIAResult
* Found a bug in HSJ, created issue #302 
* Updated the metadata creation to only include init params for optimizer and loss (model_handler was breaking in unit tests)

## Resolved Issues

- [ ] fixes #282 
- [ ] fixes #287 
- [ ] fixes #288
- [ ] fixes #290 
- [ ] fixes #291 

## How Has This Been Tested?
* All attacks have been run and ensured to generate results.
* Unit tests pass


henrikfo (approved):

leakpro/reporting/report_utils.py
```diff
@@ -0,0 +1,100 @@
...
+
+    config = dict(sorted(config.items()))
+
+    exclude = ["attack_data_dir"]
+
+    config_name = ""
+    for key, value in zip(list(config.keys()), list(config.values())):
+        if key in exclude:
+            pass
+        elif type(value) is bool:
```
fazelehh, line 29: why? does it mean to overwrite key: False and key: True?

fazelehh (approved):

PR #309, opened 2025-04-15
Audit multiple instances of same attack

### LINKED ISSUE #293: Allow multiple of the same attack in one audit
# Feature
Right now running two or more instances of the same attack with different parameters is not possible in audit.yaml.
This is because attack name is used as key in yaml file and yaml does not allow duplicate keys.
This could possibly be fixed by changing audit.yaml structure and making attack_list an actual yaml list of attack objects.

### LINKED ISSUE #307: Allow loading MIA results
# Feature

### LINKED ISSUE #308: Store attacks when they are done instead of after all attacks are done
# Feature

johanos1:
# Description

Summary of changes

* LeakPro can now audit multiple instance of the same attack. 
   - Achieved by converting the audit config to a list. The schema and audit scheduler are changed accordingly.
* We now support loading previous MIAResults based on a hash between the target model and the attack config.

## Resolved Issues

- [ ] fixes #293 
- [ ] fixes #307
- [ ] fixes #308

## How Has This Been Tested?
* CIFAR10 example pass when RMIA is run twice with different configs
* If one attack has been run, it will be loaded on the next time attacks are run provided the target has not changed
* Unit tests are updated and pass

## Related Pull Requests

- #289


leakpro/attacks/mia_attacks/qmia.py
```diff
@@ -322,36 +322,60 @@ def run_attack(self:Self) -> MIAResult:
...
-        logger.info("Attack completed")
-
-        # pick out the in-members and out-members signals
-        self.in_member_signals = self.target_logits[self.audit_dataset["in_members"]]
-        self.out_member_signals = self.target_logits[self.audit_dataset["out_members"]]
-
-        predictions = np.less(score, self.target_logits[np.newaxis, :])
 
-        # set true labels for being in the training dataset
+        # set true labels for being in the training datasetz
```
henrikfo, line 326: dataset*, like the comment for the other attacks

leakpro/input_handler/mia_handler.py
```diff
@@ -127,11 +127,10 @@ def _load_criterion(self:Self) -> None:
         if loss_cls is None:
             raise ValueError(f"Criterion {criterion_config.name} not found in torch.nn.modules.loss")
 
-        # Retrieve only valid __init__ parameters
-        valid_params = inspect.signature(loss_cls.__init__).parameters
-        filtered_params = {k: v for k, v in criterion_config.params.items() if k in valid_params}
+        # Overwrite with parameters that are relevant to the loss function
+        self.target_model_metadata.criterion = LossConfig(name=criterion_config.name, params=criterion_config.params)
```
henrikfo, line 131: Where is the validation of params moved to?

leakpro/reporting/report_handler.py
```diff
@@ -1,174 +1,79 @@
...
-
-        self.logger.info(f"Saving results for {attack_name}")
-        attack_config = config.attack_list.get(attack_name, None)
-
-        if attack_config is None:
-            attack_config = AttackFactoryMIA.attack_classes[attack_name].get_default_attack_config()
-            config.attack_list[attack_name] = attack_config
-        result_data.save(path=self.report_dir, name=attack_name, config=config.model_dump())
-
-    def load_results(self:Self) -> None:
```
henrikfo, line 75: Are we moving away for saving the results for later compilation of report creation? If not, can we add this or a method that checks for, and then loads previously saved results below "if create_pdf:" in leakpro.py

henrikfo (approved):
Looks great! Left a few comments but feel free to merge after responding to them

leakpro/input_handler/mia_handler.py
johanos1: This is moved to the conversion. I ran into issues with the shadow model handler as these types of checks were not made there. It makes sense to me to have the check when saving instead of when loading. For loading, we need to have it in at least three places.

leakpro/reporting/report_handler.py
johanos1: Not sure I follow this. Can you elaborate a bit?

fazelehh (approved):

leakpro/reporting/report_handler.py
henrikfo: We previously decided that we should have the results saved to disk then loaded when we created any report/pdf. Now we are creating the report/pdf right after the results are created, without saving to disk and then loading. What I'm wondering is how do we handle already saved results for the purpose of creating a report?

johanos1: It is handled in the PR. We are storing it in the attack scheduler

PR #312, opened 2025-04-15
Dpsgd celebA

### LINKED ISSUE #267: hard coded dpsgd path
# Issue
## Problem Description
In LOS/mimic_GRUD_dpsgd_handler.py, the path to the .pkl file storing DPSGD properties is currently hardcoded. For long-term maintainability, this should be removed. If DPSGD is needed for training shadow models, the path should be moved to a more appropriate location. Otherwise, if supporting shadow model training with DPSGD is unnecessary, the line should be removed entirely

## What Needs to be Done
Evaluate whether using DPSGD in shadow model training improves attack performance in more attackable examples (CIFAR or Celeb HQ) and decide what makes the most sense based on the results.

### LINKED ISSUE #310: celebA handler outdated
# Issue
## Problem Description
The CelebAHQInputHandler is out dated and is returning the error:
`AttributeError: 'celebHqDataset' object has no attribute 'data'`

## Expected Behavior
Should be working.

## What Needs to be Done
Update The CelebAHQInputHandler and make it function like the CifarInputHandler

Run the celebA example

### LINKED ISSUE #322: DP-Optimizer method not supported
# Issue
## Problem Description
The SGD and Adam optimizers do not support "signal_skip_step" if not wrapped by Opacus's DPOptimizer when using BatchMemoryManager without DP.

## Expected Behavior
The ResNet18 DP-SGD model should work for celebA / cifar 10 when dpsgd flag is set to True and False

## What Needs to be Done
Create two cases that uses a common training loop.

code running with and without DP-SGD

### LINKED ISSUE #372: Opacus package
# Issue
## Problem Description
Opacus is not included as dependencies in the pyproject.toml.
This causes an error when runing the mia cifar example: target_model_class.py tries to ModuleValidator from opacus.validators. 

## Expected Behavior
No import error. 

## What Needs to be Done
opacus should be added to pyproject.toml.
I do not know if opacus is specific for a setting, for example federated learning. If this is the case the import should be made in some other way so correct optional dependencies can be installed.

Run example without import error. 
Do we have tests for imports? Shouldn't github checks catch this type of error?

### LINKED ISSUE #408: Opacus 1.5.4 breaks inplace activation support
# Issue
## Problem Description
Latest version of opacus, 1.5.4, breaks the dpsgd examples.
This is because of a change in handling inplace activation internally in opacus.

## Expected Behavior
As in version 1.5.3, we should be able to run predefined imported models, just like they do in their own examples.  

## What Needs to be Done
Support inplace activation replacement.

Run opacus with models that have inplace activations and without. 
Check their own resnet18 torch importing example.

henrikfo:
# Description

Summary of changes
- Updated the celebA example to work with new changes, schemas etc.
- Added DP-SGD friendly handler for the celebA example
- Fixed a typo in cifar_handler_dpsgd.py
- Adding a way to replace inplace internal function for model building (while opacus is broken) 

## Resolved Issues

- [x] fixes #310 
- [x] fixes #322 for the CelebA HQ example
- [x] fixes #372 
- [x] fixes #267 
- [x] fixes #408  

## Related Pull Requests
- #271 
- -


examples/mia/celebA_HQ/celebA_hq_dpsgd_handler.py
```diff
@@ -0,0 +1,319 @@
...
+        # Get dpsgd flag from the model
+        RUN_DPSGD = model.dpsgd
+
+        if RUN_DPSGD:
+            # Check if the model is compatible with DP-SGD
+            errors = ModuleValidator.validate(model, strict=False)
+            if len(errors) > 0:
+                logger.info("Model has privacy violations. Fixing...")
+                # Use ModuleValidator.fix to fix the models privacy violations
+                model = ModuleValidator.fix(model)
```
TheColdIce: This is already done in target_model_class.py
But maybe it's good to check the model again?

TheColdIce (changes requested):
Nice job! Great that the code is very similar to the cifar dpsgd example, this will make it easy to build a potential generell module for dpsgd. However, I have some concerns: 

**1. yaml files**
These needs to be cleaned up a bit!

**2. virtual_batch_size**
Why is virtual_batch_size needed? Why not only use batch_size? 

**3. learning**
When using dpsgd the model seems to not learn anything? This is the results from one run: 
Image
I changed params a bit, but I was not successful in getting good results. Maybe I just ran it with bad privacy params?  

**4. opacus**
Why have you restricted opacus to 1.5.3?  
I get an warning when run the code with this version and when I use the latest version (1.5.4) I get an error (see warning and error below).
It seems the problem is that nn.GroupNorm returns a view and/or the relu activations are an inplace operation. This could lead to gradients being miscalculated. It seems the error is in opacus and/or torch/torchvision. A fix/test would be to implement resnet from scratch and make sure the groupnorm and relu behaves as needed. Or just wait for opacus to fix the error. I ran an example of theirs with the latest version and I got the same error. **It could be that this warning/error is connected to the learning problem.** Let me know what you think! 
 
**warning:** FutureWarning: Using a non-full backward hook when the forward contains multiple autograd Nodes is deprecated and will be removed in future versions. This hook will be missing some grad_input. Please use register_full_backward_hook to get the documented behavior.
  self._maybe_warn_non_full_backward_hook(args, result, grad_fn)

**error:** RuntimeError: Output 0 of BackwardHookFunctionBackward is a view and is being modified inplace. This view was created inside a custom Function (or because an input was returned as-is) and the autograd logic to handle view+inplace would override the custom backward associated with the custom Function, leading to incorrect gradients. This behavior is forbidden. You can fix this by cloning the output of the custom Function.

examples/mia/celebA_HQ/celebA_hq_dpsgd_handler.py
henrikfo: When you say check the model here, I'm not sure I follow what that's referring to

henrikfo: For models that have the private validate function in the model creation process, this is not needed of course. But models that do not have that private method or if the user skips the method for its own models, it's good to have as a guardrail I think.

henrikfo:
> @henrikfo I looked through your latest commits ([ad15a50](https://github.com/aidotse/LeakPro/commit/ad15a50a3bfb716f33986cdc6d7e21431acf4c6f) to [aba9663](https://github.com/aidotse/LeakPro/commit/aba9663056c8c2bfd684880f75dbcdd28a074563)), nice solve with the opacus problem.

Thanks, I hopefully it will not break when they make even newer releases! 

> Regarding the hardcoded path problem, I noticed that cifar and celeb seems to have a similar problem: when I change the data paths in the train_config and the audit file I get an error that the target cant be found, leakpro is still looking for `target_dpsgd/`. I think leakpro dont assign the correct `dpsgd_metadata_path` to `CifarInputHandlerDPsgd` and `CelebAInputHandlerDPsgd` so they use the default: `./target_dpsgd/dpsgd_dic.pkl`. I wasn't able to test LOS, the code needs to be merged with main before the LOS example can run.

I updated the static path. What do you think of it now?

> Minor fix: there is some empty cells in celebA_hq_dpsgd.ipynb that can be removed.

Removed!

> Regarding the config files I do not think the dpsgd example should be in a separate folder. I would actually see that there was one notebook, one train_config file and one audit. First the user would train the model, then run the attacks, then implement a PET on the model and the run the attacks again. While this change wouldn't be to difficult I dont think this needs to be done in this PR. What do you think?

Yeah maybe you're right! I just feel that it's awkward with one train and audit .yaml for these two cases. Like virtual batch size is not used for the regular example. Maybe there is some clean way of doing it?

TheColdIce:
> I updated the static path. What do you think of it now?

Nice! Looks good.

> Yeah maybe you're right! I just feel that it's awkward with one train and audit .yaml for these two cases. Like virtual batch size is not used for the regular example. Maybe there is some clean way of doing it?

Okej, if the train and audit .yaml files are very different maybe it's better to keep it separated... I'll think about it, but I think we can merge this PR in the meantime.

TheColdIce (dismissed):

examples/mia/celebA_HQ/train_config_dpsgd.yaml
```diff
@@ -0,0 +1,19 @@
+run: # Configurations for a specific run
+run: # Configurations for a specific run
```
TheColdIce, line 1: remove duplicate line

TheColdIce (changes requested):

fazelehh (approved):

TheColdIce (approved):

PR #348, opened 2025-10-27
Add vectorized RMIA

ricupa:
# Description

* Added the vectorized RMIA in leakpro/attacks/mia_attacks/rmia.py, including helper routines, configuration flags, and logic to keep both classic and vectorized flows consistent. Also, added  in a self-test that compares classic vs vectorized outputs.

* Introduced examples/mia/cifar/vectorized_rmia.ipynb, a CIFAR walkthrough showing how to enable the vectorized RMIA attack and reuse cached logits/membership masks, complete with ROC plotting utilities. 

* Added examples/mia/cifar/audit_vectorized.yaml – new audit file that enables the vectorized RMIA fast path. This was not that important but I did not wanted to touch the already working code in the repository so created a new audit file.

### DIFF
--- a/examples/mia/cifar/audit_vectorized.yaml
+++ b/examples/mia/cifar/audit_vectorized.yaml
@@ -1,49 +1,25 @@
 audit:
+  random_seed: 1236
   attack_list:
   - attack: rmia
-    num_shadow_models: 2
-    online: false
-    training_data_fraction: 0.5
-    vec_batch_points: 1024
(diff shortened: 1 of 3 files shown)


examples/mia/tabular_mia/utils/adult_model_preparation.py
henrikfo: Why is this removed?

PR #355, opened 2025-11-14
Shuffled sorting method for neighbour distance.

### LINKED ISSUE #354: Deterministic sorting of nearest neighbours causes false positives for linkability
# Issue
## Problem Description
the argsort function in mixed....py will allways return the same order for the same array. This leads to false positive links created.

## Expected Behavior
Should never give a succesful outcome when all distances for a all categorical linability attack is 1.

## What Needs to be Done
Change the sorting function

Run a linbability evaluation of single aux columns,  both needs to be categorical, n_attacks = 1 andn_neighbours>3.
Check distances calculated and links created and matched.

henrikfo:
# Description
For the Linkbility attack, the nearest_neighbors function was returning the same array of indecies all the time, causing false positives that sometimes incorrectly showed a 100% attack success rate. 

Summary of changes
* Implemented general sorting function that is NOT deterministic.

## Resolved Issues

- [X] fixes #354 

## How Has This Been Tested?
Was tested for the specific bug and for performance.

Using LinkabilityEvaluator, with 1-10 n_neighbours (does not scale) and varying n_attack: [1, 10, 100, 1000]

old code:
[131 ms ± 3.09 ms, 140 ms ± 4.01, 187 ms ± 9.62 ms, 709 ms ± 51.5 ms]

old code with addnoise (A solution to the issue):
[131 ms ± 6.82 ms, 135 ms ± 7.12 ms, 220 ms ± 18.7 ms, 992 ms ± 65.4 ms] 

myshuffle:
[129 ms ± 2.03 ms, 128 ms ± 3.82 ms, 131 ms ± 6.74 ms, 665 ms ± 31.2 ms]

### DIFF
--- a/leakpro/tests/tests_synthetic_data_attacks/anonymeter_tests/test_mixed_types_n_neighbors.py
+++ b/leakpro/tests/tests_synthetic_data_attacks/anonymeter_tests/test_mixed_types_n_neighbors.py
@@ -100,50 +100,30 @@ def test_mixed_type_n_neighbors_shape(n_neighbors: int, n_queries: int) -> None:
     assert isinstance(ids, np.ndarray)
     assert ids.shape == (n_queries, n_neighbors)
 
-
-def test_returns_valid_permutation():
-    """Check that the output contains all indices from 0 to n-1 exactly once."""
+def test_shuffled_argsorted_logic():
+    """
+    Tests that shuffled_argsorted returns valid indices that sort the array,
+    verifying against specific valid permutations for ties.
+    """
+    # Array with duplicates to allow for multiple valid index permutations
     arr = np.array([10, 5, 8, 5, 2])
+    
     result = shuffled_argsorted(arr)
     
-    # Result should have same shape
-    assert result.shape == arr.shape
-    # Result should contain all indices (0 to 4)
-    expected_indices = np.arange(len(arr))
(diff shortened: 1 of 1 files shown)


leakpro/tests/tests_synthetic_data_attacks/anonymeter_tests/test_mixed_types_n_neighbors.py
```diff
@@ -99,3 +99,75 @@ def test_mixed_type_n_neighbors_shape(n_neighbors: int, n_queries: int) -> None:
...
+    
+    # Get the argsort
+    idx = shuffled_argsorted(arr)
+    
+    # If it were a standard stable argsort, idx would be [0, 1, 2, ... 99]
+    # Since we shuffle, it is statistically impossible for it to be perfectly ordered
+    # for a large N.
+    is_perfectly_ordered = np.all(np.diff(idx) == 1)
+    
+    assert not is_perfectly_ordered, "Indices for identical values were not shuffled (result looked stable)."
```
gpadres: i think this test is unnecessary if we include in previous test an assertion of possible combinations

gpadres (dismissed):
I made a couple of comments on the test, looks good eitherway, good job Henrik!

gpadres (approved):

PR #357, opened 2025-11-26
Added n_neighbors argument to find_links()

### LINKED ISSUE #356: Linkability attack function find_links() uses default value for n_neighbours
# Issue
## Problem Description
Linkability attack function find_links() always uses default value n_neighbours for the link creation.

## Expected Behavior
Linkability attack function find_links() should take the user defined value

## What Needs to be Done
Make the function take the n_neighbours argument

Check the code, linkability_evaluator.py, line 96 calls function on line 49 without any argument.
Put a print statement inside find_links() then run a very simple linkability attack with n_attacks=1, n_attacks=1 and vary the n_neighbours argument for the attack.

henrikfo:
# Description

Summary of changes
Added n_neighbors argument to find_links()

## Resolved Issues

- fixes #356

## How Has This Been Tested?
Same check as stated in the issue.

### DIFF
--- a/leakpro/tests/tests_synthetic_data_attacks/anonymeter_tests/test_linkability_evaluator.py
+++ b/leakpro/tests/tests_synthetic_data_attacks/anonymeter_tests/test_linkability_evaluator.py
@@ -212,7 +212,6 @@ def test_linkability_naive_attack(n_neighbors: int, e_naive_rate: float) -> None
         n_jobs=1
     )
     results = evaluator.evaluate()
-    naive_rate = results.naive_rate
     assert isinstance(results, EvaluationResults)
     naive_rate = results.naive_rate
     assert naive_rate.ci[0] <= e_naive_rate


leakpro/tests/tests_synthetic_data_attacks/anonymeter_tests/test_linkability_evaluator.py
```diff
@@ -140,24 +138,49 @@ def mock_np_random_default_rng(seed: int = None) -> Callable:
...
-    assert isinstance(results, EvaluationResults)
     main_rate = results.main_rate
-    assert main_rate.rate == e_main_rate
-    assert abs(main_rate.ci[0]-e_ci[0])<0.0001
-    assert abs(main_rate.ci[1]-e_ci[1])<0.0001
+    
+    # Rate should strictly hover at 0.5 because exactly 50% of records are linkable.
+    # The other 50% are disjoint strings that cannot be linked.
+    assert main_rate.rate == pytest.approx(e_main_rate, abs=0.05)
+    assert((main_rate.ci[0]-e_ci[0])>0.0)
```
gpadres: could we set a limit to the confidence intervals ? (e.g. before we had abs difference < 0.0001)

henrikfo: We could hard code it but e_ci[0] and main_rate.ci[0] does not follow each other for n_neighbors > 1 so (main_rate.ci[0]-e_ci[0])>0.0 is I think the best we can do.

Issue #365, opened 2026-01-27
Test for time series


fazelehh:
# Issue
## Problem Description

This is some issues regarding PR #361  : 
1. There are no tests. However, the code looks good, and there are no apparent issues with it. Can we add an Issue of missing tests to keep track of that, then implement them at a later stage?
2. No random seeds, for instance data_indices = np.random.choice(..), or train_test_split(...) in dts.py, which will make reproducability harder.
3. Might be appropriate with a warining when users sets a too-low "clf_data_fraction". I think users will understand this without the warning but it might be a good add as a failsafe for runtime, when we are trying to mature the tool. A simple # After creating MIC dataset if len(self.mic_data) < 10: logger.warning(f"MIC dataset only has {len(self.mic_data)} samples...")

PR #368, opened 2026-02-02
LDG-11: Time Series MIA

fazelehh:
## Summary

This PR adds time series MIA support: two new distance signals (DTW and MSM), configurable shadow model sampling, handler fixes, and a reworked multi-signal LiRA attack (functional signals + corrected offline scoring).

### New signals

- **DTW** (`DTW`): Dynamic Time Warping distance between model output and target series. Requires optional `sktime` dependency — raises a clear `ImportError` if not installed.
- **MSM** (`MSM`): Move-Split-Merge distance, implemented from scratch in `leakpro/signals/utils/msm.py` with no external dependencies.
- Both are registered in `SIGNAL_REGISTRY` and usable via `signals` in attack configs (e.g. `multi_signal_lira`).

### MS-LiRA: functional signals + corrected offline scoring

- **Functional-signal migration.** `multi_signal_lira` now scores **cached logits** with functional signals instead of re-querying every shadow model once per signal. The old class-based path re-ran each model `n_signals` times (issue #390); `prepare_attack` now loads each model's logits exactly once (`num_shadow_models + 1` loads total) and applies every signal to the cached array.
- **Corrected offline score.** The offline path now implements arXiv:2509.04169 Eq. (2): the one-sided multivariate Gaussian tail under the OUT model (summed per-signal `norm.logcdf`), with each signal oriented so higher = more member-like via `SIGNAL_MEMBERSHIP_DIRECTION`. Previously the offline branch reused the online log-likelihood-ratio form with `in_prs = 0`, which is not the paper's offline score. For a single signal it now reduces exactly to offline LiRA (Carlini Eq. 4).
- **Online score** factored into `_online_score` (summed per-signal LLR, Eq. 1); both score paths are static helpers and unit-tested.
- Added a functional `loss` signal.
- Undeclared signals (no membership-direction entry) now fail loudly at construction, before any models are trained.

### Config rename

- The `multi_signal_lira` config field `signal_names` is renamed to **`signals`** (user-facing). The internal resolved-callables list is `signal_fns`. Class-style names (`ModelRescaledLogits`) and functional names (`rescaled_logits`) are both accepted for back-compat.

### Configurable shadow model sampling

- Added `sampling_method` field to `ShadowModelConfig` (default: `"balanced"`).
- `create_shadow_models()` now accepts `sampling_method="balanced"` (existing behavior) or `"random"` (delegates to handler for custom sampling logic).
- Added `sample_shadow_indices()` to `AbstractInputHandler` as an overrideable method — time series handlers (e.g. `IndividualizedInputHandler`) can override it to sample by individual rather than by record.

### Handler fixes

- `MIAHandler.get_dataset()`: default `params` now uses `population.return_params()` instead of `{}`, preserving dataset-specific settings (e.g. scaler, lookback) when recreating datasets for shadow model training.
- `MIAHandler.get_optimizer()`: uses `inspect.signature()` to filter out invalid optimizer kwargs before instantiation, preventing crashes when time series configs pass extra parameters.

### Bug fix

- Fixed a `DotMap` compatibility bug in `shadow_model_handler.py` where `sampling_method` would be read as an empty `DotMap()` instead of defaulting to `"balanced"` when the key was absent from the config.

### Dependencies

- Added `sktime` to `mia` optional deps and `dev` group in `pyproject.toml`.

### Tests

- `leakpro/tests/mia_attacks/utils/test_time_series_signals.py` (18 tests): MSM cost/distance units, end-to-end MSM signal, DTW `ImportError` without sktime, registry presence, shadow sampling, `get_optimizer()` kwarg filtering.
- `leakpro/tests/mia_attacks/attacks/test_multi_signal_lira.py` (new): online/offline, single- and multi-signal runs, signals resolve to functional functions, no per-signal model re-query (load_logits called exactly `num_shadow_models + 1` times), undeclared-direction failure, offline/online scoring formulas.
- `leakpro/tests/signals/test_functional.py` (new): functional signal unit tests.

## Test plan

- [x] All existing tests pass with no regressions
- [x] New unit/integration tests pass (time series signals, MS-LiRA, functional signals)
- [x] Manual run: time series example (`examples/mia/time_series_mia`) with `multi_signal_lira` over the full signal set
- [x] Manual run: existing MIA attacks (lira, rmia) still work end-to-end with default `balanced` sampling


leakpro/attacks/mia_attacks/lira.py
```diff
@@ -67,8 +70,17 @@ def __init__(self:Self,
         for key, value in self.configs.model_dump().items():
             setattr(self, key, value)
 
+        # TODO: Remove self.shadow_models? Is it ever used? Parent classes do not require it to exist.
         self.shadow_models = []
 
+        # Map legacy class-style signal names (pre-functional refactor) to their functional equivalents
```
henrikfo: Could we move this functionality to "leakpro.signals.functional", and in lira.py use signals.get("rescaled_logits") instead?

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -143,17 +124,22 @@ def prepare_attack(self:Self)->None:
                                                                               )
 
         self.shadow_models, _ = ShadowModelHandler().get_shadow_models(self.shadow_model_indices)
-
         self.out_indices = ~ShadowModelHandler().get_in_indices_mask(self.shadow_model_indices, self.audit_dataset["data"]).T
 
-        true_labels = self.handler.get_labels(self.audit_dataset["data"])
-        self.target_logits = ShadowModelHandler().load_logits(name=f"target_{ShadowModelHandler().target_model_hash}")
+        true_labels = self.handler.get_labels(self.audit_dataset["data"]).squeeze()
+        self.target_model_logits = self.signal(
```
henrikfo: I don't see why we are changing from "target_logits" to "target_model_logits". If we are for clarity, then we should make sure this naming convention is followed by all attacks!

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -67,8 +70,17 @@ def __init__(self:Self,
         for key, value in self.configs.model_dump().items():
             setattr(self, key, value)
 
+        # TODO: Remove self.shadow_models? Is it ever used? Parent classes do not require it to exist.
         self.shadow_models = []
```
henrikfo: It's probably stale old code. We used to import the actual shadow models, and then fetch the logits from them.
If self.shadow_models is not used, it should be safe to remove.

leakpro/attacks/mia_attacks/multi_signal_lira.py
```diff
@@ -76,14 +127,23 @@ def __init__(self:Self,
                     There is no data left for the shadow models.")
 
         self.shadow_models = []
```
henrikfo: Might be stale and safe to remove here as well.

leakpro/attacks/mia_attacks/lira.py
```diff
@@ -143,17 +124,22 @@ def prepare_attack(self:Self)->None:
...
         for indx in self.shadow_model_indices:
-            self.shadow_models_logits.append(ShadowModelHandler().load_logits(indx=indx))
-
-        self.shadow_models_logits = np.array([self.rescale_logits(x, true_labels) for x in self.shadow_models_logits])
-        self.target_logits = self.rescale_logits(self.target_logits, true_labels)
+            self.shadow_models_logits.append(
+                self.signal(
+                    ShadowModelHandler().load_logits(indx=indx),
+                    true_labels
+                ).T
```
henrikfo: Does this still have the same behavior in the loop on line 201?

examples/mia/time_series_mia/audit.yaml
```diff
@@ -0,0 +1,60 @@
...
+      signals: ["mse", "mae", "smape", "rescaled_smape", "seasonality", "trend", "ts2vec", "dtw", "msm"]
+      num_shadow_models: 2
+      online: True
+
+    # LIRA
+    - attack: lira
+      signal: msm  # mse, mae, smape, rescaled_smape, seasonality, trend, ts2vec, dtw, msm
+      num_shadow_models: 2
+      online: True  # True, False
+      individual_mia: False  # True, False
```
henrikfo: "individual_mia" is not used by lira.py

henrikfo: Add extra="forbid" to lira.py to prevent similar silently wrong arguments.

examples/mia/time_series_mia/main.py
```diff
@@ -5,7 +5,7 @@
 import os, sys, yaml, numpy as np, matplotlib.pyplot as plt, torch, pandas as pd
 
 project_root = os.path.abspath(os.path.join(os.getcwd(), "../../.."))
```
henrikfo: Not important but good to know, this make the main.py a executable that uses a relative path, meaning someone needs to be in the same folder. This is not a problem in notebooks since the kernel is connected to the the notebook and therefore in the same relative path

examples/mia/time_series_mia/run_audit.py
```diff
@@ -0,0 +1,25 @@
+import os
+import sys
+
+project_root = os.path.abspath(os.path.join(os.getcwd(), "../../.."))
```
henrikfo: Same here, the executable uses a relative path, meaning someone needs to be in the same folder.

examples/mia/time_series_mia/main.py
henrikfo: Is not used, and the save directory is overwritten with with "target/"

leakpro/attacks/mia_attacks/rmia.py
```diff
@@ -96,6 +114,27 @@ def __init__(self:Self,
         self.shadow_models = None
         self.shadow_model_indices = None
 
+        # Resolve the residual-likelihood family from the training criterion: each regression loss
+        # implies a noise model (MSE -> Gaussian, L1 -> Laplace, SmoothL1/Huber -> Huber), and the
+        # signal probability must use the matching likelihood. Classification losses keep the
+        # softmax probability of the true class.
+        criterion = self.handler.get_criterion()
```
henrikfo, line 121: data_handler.py always return MSELoss(). 
The "loss: MSE" in train_config.yaml is redundant, if "loss: MAE" then its wrong and it will select the wrong "family_by_criterion"

leakpro/attacks/mia_attacks/lira.py
henrikfo: Also, this might become a problem if we are implementing multi-dimensional logits/signals.

leakpro/attacks/mia_attacks/multi_signal_lira.py
```diff
@@ -76,14 +127,23 @@ def __init__(self:Self,
...
-        self.signals = [create_signal_instance(signal_name) for signal_name in self.signal_names]
+
+        # Membership direction per signal (+1 higher=member, -1 lower=member). Resolved first so
+        # an unknown/undeclared signal fails with the direction error before any models are trained.
+        self.signal_directions = np.array(
+            [_signal_membership_direction(name) for name in self.signals]
+        )
+
+        # Functional signals run on cached logits (no per-model re-query); see prepare_attack.
+        self.signal_fns = [_resolve_signal_fn(name) for name in self.signals]
```
henrikfo: signals cannot be left empty, signals=[], should just have a small check for that.

leakpro/attacks/mia_attacks/multi_signal_lira.py
henrikfo: Change "ge=0.0" to "gt=0.0", to not allow std of 0.0

henrikfo: (It's not really part of this PR but since we are allowing user to set std_eps, then we should probably not allow std_eps = 0.0)

leakpro/attacks/mia_attacks/multi_signal_lira.py
```diff
@@ -129,14 +189,45 @@ def prepare_attack(self:Self) -> None:
...
+        # series matching the (already squeezed) cached logits).
+        targets = np.asarray(self.handler.get_labels(audit_indices)).squeeze()
+        target_logits = ShadowModelHandler().load_logits(name="target")
+        shadow_logits = [ShadowModelHandler().load_logits(indx=indx) for indx in self.shadow_model_indices]
+
+        shadow_models_signals = []
+        target_model_signals = []
+        for signal_fn, signal_name in zip(self.signal_fns, self.signals):
+            logger.info(f"Calculating {signal_name} for the target model")
+            target_model_signals.append(signal_fn(target_logits, targets))
```
henrikfo: when the signal_fn is "ts2vec" this line trains a encoder, which is not returned and reused for the shadow model logits. Instead it trains a new instance, so the encoders used here inside the ts2vec signal is two separate ones. I'm might be wrong but I think the paper uses the one and the same encoder for both, so we need to return the first encoder and use it in the second set of logits, or vice-versa, or train a encoder the is forwarded to both signal instances.

leakpro/attacks/mia_attacks/multi_signal_lira.py
```diff
@@ -129,14 +189,45 @@ def prepare_attack(self:Self) -> None:
...
+
+            logger.info(f"Calculating {signal_name} for all {self.num_shadow_models} shadow models")
+            # (n_shadow_models, n_audit_points) -> (n_audit_points, n_shadow_models)
+            per_model = np.array([signal_fn(logits, targets) for logits in shadow_logits])
+            shadow_models_signals.append(per_model.T)
```
henrikfo, line 173: The transpose, ".T", might be incompatible in-between all different signals. 
For MS-LIRA, the signal would return one set of sizes (shadow_models, samples, horizon, variables), 
(shadow_models, samples, horizon, variables).T  =>  (variables, horizon, samples, shadow_models) then
then np.stack(shadow_models_signals, axis=-1) stacks on the samples dimension.
From what I can tell the MSE and DTW is not compatible with that stacking.

Can we check that this works fine? 
Could be verified with selecting all signals in one minimal attack run.

leakpro/attacks/mia_attacks/rmia.py
```diff
@@ -96,6 +114,27 @@ def __init__(self:Self,
         self.shadow_models = None
         self.shadow_model_indices = None
 
+        # Resolve the residual-likelihood family from the training criterion: each regression loss
```
henrikfo, line 117: It this the equivalent to the papers g(s) = 1/(1 + s) transformation?

leakpro/attacks/mia_attacks/rmia.py
```diff
@@ -35,11 +46,18 @@ class AttackConfig(BaseModel):
                                               description="Part of available attack data to use for shadow models")
         online: bool = Field(default=False,
                              description="Online vs offline attack")
+        attack_data_fraction: float = Field(default=0.5,
```
henrikfo, line 49: Should we add an alias for "z_data_sample_fraction" to "attack_data_fraction"? Otherwise we must change all the other audit.yaml - rmia configs too! Since this would break the audit.yaml files right?

leakpro/signals/functional.py
```diff
@@ -0,0 +1,207 @@
...
+    Returns:
+        np.ndarray: The logits unchanged.
+
+    """
+    _ = targets  # not used, present here for function consistency by convention.
+    return logits
+
+
+def rescaled_logits(logits: np.ndarray, true_labels: np.ndarray) -> np.ndarray:
+    """Rescale the logits to a range of [0, 1].
```
henrikfo: Its more like f_rescaled: X -> Y, where X is [0, 1] and Y is R = (-inf, inf)

leakpro/signals/functional.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def rescaled_smape(logits: np.ndarray, targets: np.ndarray) -> np.ndarray:
+    """Per-point SMAPE mapped through a logit transform to spread values over the real line."""
+    assert logits.shape == targets.shape
+    numerator = np.abs(logits - targets)
+    denominator = np.abs(logits) + np.abs(targets) + 1e-30
+    fraction = numerator / denominator
+    smape_loss = np.mean(fraction, axis=tuple(range(1, logits.ndim)))
+    return np.log(smape_loss / (1 - smape_loss + 1e-30))
```
henrikfo: maybe add a 1e-30 in the numerator too? or clip it? there's a possibility that you have a perfect prediction and then smape_loss = 0 => np.log(0) = -inf
Because later in MS-lira, signals are averaged and a single -inf will ruin it.

leakpro/signals/functional.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def ts2vec(logits: np.ndarray, targets: np.ndarray, batch_size: int = 256) -> np.ndarray:
+    """Per-point distance between TS2Vec embeddings of logits and targets."""
+    assert logits.shape == targets.shape
+    if logits.ndim == 2:  # expand dims if times series are univariate
+        logits = np.expand_dims(logits, axis=2)
+        targets = np.expand_dims(targets, axis=2)
+    device = "cuda:0" if cuda.is_available() else "cpu"
+    torch.backends.cudnn.deterministic = False
```
henrikfo: this is a global torch operation, not just inside the function scope. You could save the state before its needed, then set to False, then restore the torch state.

leakpro/signals/utils/msm.py
```diff
@@ -0,0 +1,44 @@
+"""Move-Split-Merge distance for multivariate time-series."""
+
+import numpy as np
+
+
+def mv_msm_cost(a: np.ndarray, b: np.ndarray, c: np.ndarray, cost: float = 1.0) -> np.ndarray:
```
henrikfo, line 6: There is a actual multivariate MSM, when each time point is not a scalar but a vector.
I don't know if the multivariate one was neglected in favor of the univariate one, but it seems to me that the multivariate one should be the version to implement based on the mv_msm_distance() arguments; "First time series of shape (n, num_variables)."

Never the less; """The implementation is consistent with standard MSM when the input is univariate. For multivariate inputs, however, it applies the univariate MSM cost independently to each coordinate and sums those costs while using one shared alignment path. This is not published independent multivariate MSM, which runs the complete MSM algorithm separately for each variable and then combines the final distances. It is also not published dependent multivariate MSM, which treats each timestamp as one vector-valued point, uses squared Euclidean distance for move operations, and determines whether a vector lies “between” two other vectors using a hypersphere-based test.""" - my computer friend :)

henrikfo (changes requested):

examples/mia/time_series_mia/main.ipynb
```diff
@@ -12,7 +12,27 @@
    "execution_count": null,
```
henrikfo, line 12: <pre class="ql-syntax" spellcheck="false">Line #9.    from examples.mia.time_series_mia.utils.models.LSTM import LSTM
</pre><p>Nothing important, but maybe we should import all models on one line or just choose one model to use</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aidotse/LeakPro/pull/368/discussion/#comment-3771418558'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='1'/>

henrikfo, line 12: <pre class="ql-syntax" spellcheck="false">Line #5.    if model_name == "LSTM":
</pre><p>Nothing important, by should we just choose one model for the example?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aidotse/LeakPro/pull/368/discussion/#comment-3771418565'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='6'/>

leakpro/attacks/mia_attacks/lira.py
TheColdIce: Done. Resolution moved to leakpro.signals.functional as get(name), plus direction(name) for
the membership orientation. Both use explicit registries instead of getattr on the module
namespace. MS-LiRA had its own copy of this logic, so that is deleted and both attacks share
one alias table now.

Turned out the getattr lookup returned wrong objects, not just bad error messages:
"signal: TS2Vec" in a LiRA config resolved to the imported TS2Vec model class instead of the
ts2vec signal function, since hasattr also matches the module's imports. Regression test added.

Side effect: LiRA and MS-LiRA now accept the same signal names. "signal: MSE" used to work in
MS-LiRA and fail in LiRA.

leakpro/attacks/mia_attacks/lira.py
TheColdIce: The rename was to mach MS-LiRA. It used both taget_model_logits and  target_logits. I have renamed the variable to target_signals in LiRA and MS-LiRA (also sample_target_signals in MS-LiRA), since this is a more correct naming. 

Regarding naming convention over all attacks, you already have different names for similar variables:
- base: logits_target
- attack_p: attack_signal/audit_signal
- loss traj: target
- ...

I don't have a super-strong opinion on the naming, as long as it is clear what the variable is. Please let me know what you think we should do.

leakpro/attacks/mia_attacks/lira.py
TheColdIce: Removed. Also, the two test assertions in test_lira.py that used it now check shadow_model_indices instead.

leakpro/attacks/mia_attacks/multi_signal_lira.py
TheColdIce: Removed.

leakpro/attacks/mia_attacks/lira.py
TheColdIce: Same behavior, yes, the .T was a no-op: numpy .T on a 1-D array returns it unchanged,
so the stack came out (n_models, n_points) either way. Removed. It came from MS-LiRA,
where the transpose acts on a real 2-D array.

Your second point is already broken today though. "logits" is a selectable signal that
returns its input unchanged, so on a 4-class model it is (n_points, 4), and run_attack
read n_audit_samples off the class axis. Removing the .T alone does not fix it, since
LiRA fits a scalar Gaussian per point.

So LiRA and MS-LiRA now check that a signal returns one value per audit point and raise
naming the signal. 11 of 12 signals pass, only "logits" is rejected.

examples/mia/time_series_mia/main.py
TheColdIce: Removing main.py, just a notebook is cleaner.

examples/mia/time_series_mia/run_audit.py
TheColdIce: Also removing run_audit.py.

examples/mia/time_series_mia/main.py
TheColdIce: Removed main.py (See comment above).

examples/mia/time_series_mia/audit.yaml
TheColdIce: Fixed.

leakpro/attacks/mia_attacks/rmia.py
TheColdIce: Good catch. get_criterion() now reads train.loss and resolves it through the same
mapping create_trained_model_and_metadata uses, factored into one function so the
handler and the training path cannot drift apart again.

It turned out to affect the shadow models too, not just RMIA: they are built via
get_target_replica(), so they were trained with MSELoss while the target used L1Loss,
and their metadata still said 'l1loss' because that name is copied from the target.
The same fix covers it.

Filed #439 for the general gap, that nothing verifies get_criterion() against the
criterion recorded for the target.

leakpro/attacks/mia_attacks/multi_signal_lira.py
TheColdIce: Added a pydantic Field with min_length=1.

leakpro/attacks/mia_attacks/multi_signal_lira.py
TheColdIce: Fixed.

leakpro/attacks/mia_attacks/multi_signal_lira.py
TheColdIce: Checked with all 9 signals, using the cached logits from a real run of this example
(1190480 x 20) rather than synthetic data. Took 200 points and all 6 shadow models and
pushed every signal through the same .T and np.stack path:

    all 9 signals -> (n_points,)
    stacked shadow (200, 6, 9)   = (n_points, n_models, n_signals)
    stacked target (200, 9)      = (n_points, n_signals)

Also ran it with a synthetic multivariate input (200, 20, 3), since this example is
univariate and cannot exercise the variables axis your case turns on. Same result.

The 4-D case cannot arise because every signal reduces to one scalar per point: mse,
mae and smape average over all trailing axes, trend and seasonality take a norm over
both, dtw, msm and ts2vec return one value per sample. So per_model is always
(n_models, n_points) and .T is a plain 2-D transpose. MSE and DTW are fine.

You were right that nothing enforced it though. Same hole as your lira.py comment, so
MS-LiRA now checks that each signal returns exactly (n_audit_points,) for the target
and every shadow model, and raises naming the signal otherwise.

A test running all signals in one attack still needs a time series input handler
fixture, which is #365.

leakpro/attacks/mia_attacks/rmia.py
TheColdIce: No, and deliberately not. The code uses the likelihood implied by the training
criterion, exp(-MSE/2sigma^2) for MSE, exp(-MAE/b) for L1, the Huber density for
Huber/SmoothL1, with the scale estimated from shadow-model residuals.

Both mappings are monotone decreasing, so for ranking one model they are
interchangeable, but RMIA takes a ratio p(x|theta)/p(x) and there the form matters.
g(s) = 1/(1+s) has no scale parameter, so it is unit dependent: at the error
magnitudes of a scaled forecasting target it maps everything to about 0.999 and the
ratios collapse toward 1, and rescaling the data changes the attack output. For two
points with errors 0.001 and 0.002 the ratio is 1.001 under g, and 1.105 under the
Gaussian surrogate with a shadow-estimated scale.

I think it is the better choice: RMIA needs Pr(x|theta), and for a regression target
the analogue of the softmax probability of the true class is the density under the
noise model the loss implies. Happy to document the deviation from the paper in the
docstring if you want it recorded there.

leakpro/attacks/mia_attacks/rmia.py
TheColdIce: No changes needed, I checked both this branch and main. z_data_sample_fraction only
ever appears in rmia.py and its tests, never in a config: all eight rmia blocks in
main's example audit.yaml files set only online, num_shadow_models,
training_data_fraction or gamma, so they all relied on the 0.5 default. The rename in
95430b28 updated the three places that referenced it, and the time_series example was
written with the new name.

So an alias would only cover a user outside the repo who set a key that none of our
own examples set, which means adding it by hand from the field description. I left it
out, but say the word if you would rather have it.

The naming is also the reason for the rename: attack_p and HSJ already call the same
concept attack_data_fraction, rmia was the odd one out.

leakpro/signals/functional.py
TheColdIce: Good catch. Changed the docstring. Also named it as Carlini's phi from LiRA Sec. IV-B, since the point of the transform is to make LiRA's per-example Gaussian fit reasonable.

leakpro/signals/functional.py
TheColdIce: Fixed, epsilon in the numerator too: np.log((smape_loss + 1e-30) / (1 - smape_loss + 1e-30)). The upper end was already guarded, so both ends are now symmetric and the output is bounded to about Â±69.

Same change applied to RescaledSMAPE in signal.py, which has a copy of the expression. Tests cover both ends.

leakpro/signals/functional.py
TheColdIce: Removed the line instead of restoring it. It only had an effect because seed_everything sets cudnn.deterministic = True, so it cancelled our own seeding, and since it never restored, everything cuDNN after the first ts2vec signal ran non-deterministically. That policy belongs in one place at startup.

Same line removed from signals/utils/get_TS2Vec.py. Verified on GPU that TS2Vec fits fine with determinism on, and the flag now survives both paths.

leakpro/signals/utils/msm.py
TheColdIce: Changed to standard univariate MSM per variable, each with its own alignment path, summed over variables.

examples/mia/time_series_mia/main.ipynb
TheColdIce: Fixed.

TheColdIce: Fixed.

leakpro/attacks/mia_attacks/multi_signal_lira.py
TheColdIce: You're right, and it was N+1 encoders, not two: the signal is called once per shadow model plus once for the target, and each call fitted its own. Same bug in lira.py.

Fixed by fitting one encoder up front and binding it into the signal, using the existing get_ts2vec_model, so it is fitted on the shadow population and hits the disk cache instead of being refitted on the audited points.

leakpro/attacks/mia_attacks/lira.py
henrikfo: Oh I see, I thought all "target_logits" was renamed to "taget_model_logits". 

I don't really have a preference either, I was under the impression that all "target logits" was named target_logits but I guess you are right, there is no set convention yet! And that would be outside the scope of the PR! 

Made this into an issue, for a another PR

leakpro/attacks/mia_attacks/rmia.py
henrikfo: Great! Just wanted to know if it was a mistake or a deliberate choice! :) 
A docstring would be awesome to have, just to make sure that it's clear for any user! :)

leakpro/attacks/mia_attacks/rmia.py
henrikfo: Okay, I thought the use was broader than that! :) Nothing to change then!

henrikfo (approved):
Great, ready to merge! :)

Issue #373, opened 2026-02-17
ShadowModelHandler.construct_balanced_assignments() breaks seqMIA


henrikfo:
# Issue
## Problem Description
The description of ShadowModelHandler.construct_balanced_assignments() does not support an uneven splits of datasets.

## Expected Behavior
If only one shadow model is to be created, as seqMIA is coded to do, it should not throw an error but also 

## What Needs to be Done
Make n = 1 pass.

## How Can It Be Tested or Reproduced
Run a MIA example and use the seqmia attack.

PR #375, opened 2026-02-18
Adding e2e testing for all mia attacks

### LINKED ISSUE #296: Thorough LeakPro runtime test.
# Feature
For the long run we should have a larger test for the whole leakpro repo. Have a deterministic minimal-runtime-example that produces some expected results, from setting up a config and to the end having a report created, a End-to-End test.

When re-/implementing various structures and process, single tests might pass, but in a larger context something inbetween tests might be changed and causing functions that have passed to now fail if run sequentially.

## Why does this matter
Right now we are rapidly implementing and refactoring the repository to attend to new features, bugs and needs, which is of course very important. By having a larger runtime test for the whole repo from start to end, E2E, we can then much easier do two things; Find new and update existing tests AND Version control.

The first, finding new tests, is self-explanatory. If the E2E test fails while all the unit test passes there could be, amongs many possibilities, two important issues. No coverage for that specific bug/feature or the unit tests needs to be updated.

The second, version control, will allow us during development to tag, or with another method, release stable versions the repo and its examples

## What Needs to Be Done
Create a minimal-deterministic test example with exact and reproducable results, tracing the whole repo.

henrikfo:
# Description

Adding e2e testing for all mia attacks. 
Making sure that if any attack breaking feature is introduced, it should be caught.

loss_traj and HSJ are failing, but these issues will be dealt with. 
- [x] HSJ will in this PR get an updated MIAResults output
- [x] loss_traj breaks because of #373 but will be fixed with #360 

- [x] Create a test that check all existing attacks, making sure that no attack is integrated without a test

## Resolved Issues
- fixes #296 

## How Has This Been Tested?
Ran the tests that is implemented here.


fazelehh (changes requested):
1. PATCHED_ATTACK_BLOCKERS is empty but the meta-test doesn't warn about it

PATCHED_ATTACK_BLOCKERS = {}  # empty
assert all(len(reason) > 0 for reason in PATCHED_ATTACK_BLOCKERS.values())  # vacuously True
This is fine logically, but if someone adds an attack to PATCHED_ATTACKS without a reason, the assertion silently passes. A comment here would help future contributors.

2. seqmia special-case sets input_size=5 with a comment about "5 metrics per trajectory step" , but TinyTimeSeriesTargetModel.input_size=2. Worth confirming seqmia uses the image path (not time-series) and that input_size in its AttackConfig is independent of the model input size.

3. result.id assumed to exist
result_file = output_dir / "results" / result.id / "result.txt"
This will fail with an AttributeError if MIAResult doesn't have an id field. Worth adding an assertion like assert hasattr(result, 'id') before using it, or confirming MIAResult always has this.

4.(optional) RMIA uses full population for train+test. A comment explaining why RMIA gets a different split would help reviewers.

henrikfo:
> 1. PATCHED_ATTACK_BLOCKERS is empty but the meta-test doesn't warn about it
> 
> PATCHED_ATTACK_BLOCKERS = {} # empty assert all(len(reason) > 0 for reason in PATCHED_ATTACK_BLOCKERS.values()) # vacuously True This is fine logically, but if someone adds an attack to PATCHED_ATTACKS without a reason, the assertion silently passes. A comment here would help future contributors.

Removed the "Patched" logic. It's unnecessary all implemented attack now runs the real attack. Might Implement something like this for an e2e for minva instead, where there might exist blockers or hard compute constraints than mia.

> 2. seqmia special-case sets input_size=5 with a comment about "5 metrics per trajectory step" , but TinyTimeSeriesTargetModel.input_size=2. Worth confirming seqmia uses the image path (not time-series) and that input_size in its AttackConfig is independent of the model input size.

seqmia uses TinyImageInputHandler and not TinyTimeSeriesTargetModel since the LSTM model in the attack is evaluating the loss trajectory and not the data itself.

> 3. result.id assumed to exist
>    result_file = output_dir / "results" / result.id / "result.txt"
>    This will fail with an AttributeError if MIAResult doesn't have an id field. Worth adding an assertion like assert hasattr(result, 'id') before using it, or confirming MIAResult always has this.
 
Fixed! Added a few statements that checks this and makes sure that MIAResult.id is a string exists etc.

> 4.(optional) RMIA uses full population for train+test. A comment explaining why RMIA gets a different split would help reviewers.

Added comments mentioning the fact that it does sample from the full population. But I assume that you meant why RMIA does sample like this and not like the paper or? Because from what I remember, RMIA could not have train + test sizes = 1. Why was this changed, do you know? I have not really followed that change and now the z, the known-out data is sampled from the whole population

leakpro/attacks/mia_attacks/HSJ.py
fazelehh: verbose is not a field in AttackHopSkipJump.AttackConfig. Pydantic silently drops it when the test sets candidate["verbose"] = False (line 392), so model_dump() never includes it. The result: HSJ always runs with verbose=True regardless of what the test or user passes.

We should add verbose: bool = Field(default=True, ...) to AttackConfig, then the loop at lines 82–83 sets it automatically, and line 93 can be deleted.

leakpro/tests/mia_attacks/attacks/test_all_attacks_end_to_end.py
```diff
@@ -0,0 +1,597 @@
...
+            assert hasattr(result, "id")
+            assert isinstance(result.id, str)
+            assert result.id
+
+            output_dir = Path(leakpro.handler.configs.audit.output_dir)
+            result_file = output_dir / "results" / result.id / "result.txt"
+            assert result_file.exists()
+            assert (output_dir / "data_objects").exists()
+
+        assert temp_root is not None
```
fazelehh: Minor: 
 line 561 initializes temp_root = None, but line 564 sets temp_root = Path(temp_dir) as the very first statement inside the with block. Line 594 is reached only if the entire with block completed without exception, which means line 564 already ran. So temp_root is never None at line 594 — the check is always true and catches nothing.

leakpro/attacks/mia_attacks/HSJ.py
fazelehh: wrong decstring. It is HSJ attack not RMIA

fazelehh (changes requested):

henrikfo:
Made an additional test in [27a3931](https://github.com/aidotse/LeakPro/pull/375/commits/27a3931de9d9dbeae6cd0e2c054ba4e4989cf870)[27a3931](https://github.com/aidotse/LeakPro/pull/375/commits/27a3931de9d9dbeae6cd0e2c054ba4e4989cf870)

Could you have a quick look at that aswell? Just a few lines of code! :)

fazelehh (approved):

PR #377, opened 2026-02-20
Oslo fix

fazelehh:
# Description

Addresses reviewer feedback on the OSLO (One-Shot Label-Only) membership inference attack implementation.

### `leakpro/attacks/mia_attacks/oslo.py`

- **Fix `min_threshold` validation** — Changed `ge=0.0` → `gt=0.0`. A value of `0` causes `np.log(0) = -inf` when computing thresholds, silently corrupting the entire threshold array.
- **Guard against `n_audits` overflow** — Added a `ValueError` before `np.random.choice` when `n_audits` exceeds the available attack data pool, giving a clear error instead of a cryptic numpy crash.
- **Refactor duplicated loss computation** — `_optimization_objective` and `_stop_criterion` were identical except for which model list they iterated. Extracted a shared `_compute_loss(x0, dx, y, models)` helper to eliminate the duplication.
- **Fix label alignment in `run_attack`** — Added `shuffle=False` to the dataloader. Without it, data order could differ from `true_labels`, producing incorrect membership inference results.
- **Compact model setup loops** — Merged two identical for-loops (one for `source_models`, one for `validation_models`) into a single loop.
- **Fix `signal_values` type bug** — `predictions[:, -1]` is a boolean array; `MIAResult.from_full_scores` calls `np.argsort(-signal_values)` internally which fails on boolean arrays in newer NumPy. Added `.astype(float)` before passing.

### `leakpro/tests/mia_attacks/attacks/test_oslo.py` *(new file)*

Added 4 unit tests following the pattern of `test_lira.py` and `test_rmia.py`:

| Test | What it checks |
|---|---|
| `test_oslo_setup` | Initialization with custom configs, `description()` output, `min_threshold=0` raises |
| `test_oslo_prepare_attack` | Correct shadow model counts, audit index count, no overlap between audit and attack data |
| `test_oslo_n_audits_guard` | `ValueError` raised when `n_audits` exceeds available data |
| `test_oslo_run_attack` | Full end-to-end run returns a valid `MIAResult` with correct shape and no NaN signal values |

Tests use a minimal `FAST_OSLO_PARAMS` config (`n_audits=2`, `num_sub_procedures=1`, `num_iterations=1`, 1 source model, 1 validation model) to keep runtime low.

## Resolved Issues

- Fixes all 7 inline comments from reviewer `henrikfo`
- Note: the `ShadowModelHandler` equal-model-count limitation (reviewer comment on line 26) is a known issue being patched in a separate PR

## How Has This Been Tested?

```bash
pytest leakpro/tests/mia_attacks/attacks/test_oslo.py -v  # 4/4 passed
pytest leakpro/tests/mia_attacks/ -v                       # 21/21 passed

### DIFF
--- a/leakpro/tests/mia_attacks/attacks/test_oslo.py
+++ b/leakpro/tests/mia_attacks/attacks/test_oslo.py
@@ -8,36 +8,27 @@ from leakpro.tests.constants import get_shadow_model_config
 from leakpro.tests.input_handler.image_input_handler import ImageInputHandler
 
 FAST_OSLO_PARAMS = {
-    "n_audits": 3,
+    "n_audits": 2,
     "num_source_models": 1,
     "num_validation_models": 1,
     "num_sub_procedures": 1,
     "num_iterations": 1,
     "n_thresholds": 2,
-    "online": False,
+    "online": True,
     "training_data_fraction": 0.5,
 }
 
 
 def test_oslo_setup(image_handler: ImageInputHandler) -> None:
-    """Test that AttackOSLO initializes correctly with default and custom configs."""
-    # Default params
-    oslo_default = AttackOSLO(image_handler, None)
-    assert oslo_default is not None
-    assert oslo_default.target_model is not None
-    assert oslo_default.num_source_models == 9
-    assert oslo_default.num_validation_models == 3
-    assert oslo_default.n_audits == 500
-    assert oslo_default.online == False
-
-    # Custom fast params
-    oslo_custom = AttackOSLO(image_handler, FAST_OSLO_PARAMS)
-    assert oslo_custom is not None
+    """Test that AttackOSLO initializes correctly with custom configs."""
+    oslo_obj = AttackOSLO(image_handler, FAST_OSLO_PARAMS)
+    assert oslo_obj is not None
+    assert oslo_obj.target_model is not None
     for key, value in FAST_OSLO_PARAMS.items():
-        assert getattr(oslo_custom, key) == value
+        assert getattr(oslo_obj, key) == value
 
     # Description
-    desc = oslo_custom.description()
+    desc = oslo_obj.description()
     assert len(desc) == 4
 
     # min_threshold must be > 0 (gt=0.0)

### CODE REVIEW

File: leakpro/attacks/mia_attacks/oslo.py [hunk 1]
```diff
@@ -0,0 +1,238 @@
...
+        dx: Tensor,
+        y: Tensor
+    ) -> float:
+        if self.binary_output:
+            loss = -torch.cat([self.loss(shadow_model.model_obj(x0 + dx)[0],y) for shadow_model in self.source_models], dim = 0)
+        else:
+            loss = -torch.cat([self.loss(shadow_model.model_obj(x0 + dx),y) for shadow_model in self.source_models], dim = 0)
+        return loss.sum()
+
+    def _stop_criterion(
```
henrikfo: Code duplication of line 122, only difference is the self.source_models vs self.validation_models. Maybe we create one function that have an input argument of for these models instead

leakpro/attacks/mia_attacks/oslo.py
```diff
@@ -0,0 +1,238 @@
...
+            predictions.append(batch_predictions)
+
+        predictions = np.array(predictions).reshape(-1, self.n_thresholds)
+        signal_values = predictions[:, -1].copy().reshape(-1, 1)
+
+        logger.info(f"Accuracy: {np.sum(predictions == true_labels[:,np.newaxis], axis=0)/predictions.shape[0]}")
+        logger.info(f"TPR: {np.sum(predictions * true_labels[:,np.newaxis], axis=0)/np.sum(true_labels)}")
+        logger.info(f"FPR: {np.sum(predictions * (1 - true_labels[:,np.newaxis]), axis=0)/np.sum(1 - true_labels)}")
+
+        return MIAResult.from_full_scores(
```
henrikfo: Was able to run the notebook, but if fails on the result creation. 
line 148, `sorted_indices = np.argsort(-self.signal_values)`

Does it break for you aswell?

PR #387, opened 2026-03-04
Partial shadow inheritance

### LINKED ISSUE #313: Partial shadow model inheritance
# Feature
Allow for partial inheritance when creating shadow model instead of only allowing complete or no inheritance.

## What Needs to Be Done
modify ModelHandler to iteratively step through, check and inherit configs from the target model when not given by the user. 

## How Can It Be Tested
Pass test and not breaking existing functionality.

### LINKED ISSUE #347: Expand hashing functionality
# Feature
Expand the hashing functionality to make leakpro have fewer unintentional erroneous processes, e.g. reusing old models and objects when not intended. 
_filter() function in shadowmodelhandler should support hashing functionality, And not only account for "training_data_fraction" and "model_class" when deciding if to retrain a shadow model.

## What Needs to Be Done
Implement hashing support and expand meta_check_values.

## How Can It Be Tested
Tests should be constructed to vary parameters that effect auditing objects and results such as target models, shadow models etc. and find that leakpro will not reuse previously create objects that is no longer valid.

henrikfo:
# Description

Summary of changes
Updated the shadow model inheritance logic. 
- [x] Partial shadow_model config inheritance was implemented so missing shadow settings now fall back to the target model setup instead of requiring a fully duplicated shadow config. That includes inherited model path/class, init params, optimizer, criterion, epochs, and batch size, while still allowing shadow-specific overrides.

Furthermore, silent faults or possibility of running leakpro with unintended setups have been reduced.  
- [x] Shadow-model cache reuse was fixed so existing shadow models are no longer reused just because they share num_train and model_class. Reuse now checks the full effective training setup, including optimizer settings, criterion settings, init params, batch size, online/offline mode
- [x] Shadow_model.batch_size now actually affects training. Before the field could be present in config but be silently ignored during dataloader creation
- [x] Shadow metadata was expanded so saved shadow runs now record the fields needed for safe reuse decisions, including optimizer params, criterion params, batch size, and model module path
- [x] Tests were added to cover the new behavior: one verifies partial inheritance from the target setup, one checks partial optimizer override behavior, one ensures stale cached shadow metadata is rejected when configs differ, and one verifies that the configured shadow batch size is really used

## Resolved Issues

- [x] fixes #313
- [x] fixas #347  

## How Has This Been Tested?
This was tested by running the celebA DP example with both target model and shadow models without DPsgd enabled. Then another time, setting only 
`shadow_model:`
`   init_params:`
`       dpsgd: True`

### DIFF
--- a/leakpro/attacks/utils/model_handler.py
+++ b/leakpro/attacks/utils/model_handler.py
@@ -39,6 +39,7 @@ class ModelHandler():
 
         # get the bluepring for the model
         if self.use_target_model_setup:
+            self.model_path = handler.configs.target.module_path
             self.model_class = handler.target_model_blueprint.__name__
             self.model_blueprint = handler.target_model_blueprint
         else:
@@ -51,23 +52,29 @@ class ModelHandler():
                 raise ValueError(f"Failed to create model blueprint from {self.model_class} in {self.model_path}") from e
 
         if self.use_target_model_setup:
-            self.init_params = target_setup.init_params
+            self.init_params = (target_setup.init_params or {}).copy()
             optimizer_name = target_setup.optimizer.name
-            self.optimizer_config = target_setup.optimizer.params
+            self.optimizer_config = (target_setup.optimizer.params or {}).copy()
             criterion_name = target_setup.criterion.name
-            self.loss_config = target_setup.criterion.params
+            self.loss_config = (target_setup.criterion.params or {}).copy()
             self.epochs = target_setup.epochs
+            self.batch_size = target_setup.data_loader.params.get("batch_size")
         else:
             # Inherit defaults from target and apply caller overrides when present.
-            self.init_params = target_setup.init_params.copy()
+            self.init_params = (target_setup.init_params or {}).copy()
             self.init_params.update(caller_configs.init_params or {})
             optimizer_cfg = caller_configs.optimizer or target_setup.optimizer
             criterion_cfg = caller_configs.criterion or target_setup.criterion
             optimizer_name = optimizer_cfg.name
-            self.optimizer_config = optimizer_cfg.params
+            self.optimizer_config = (optimizer_cfg.params or {}).copy()
             criterion_name = criterion_cfg.name
-            self.loss_config = criterion_cfg.params
+            self.loss_config = (criterion_cfg.params or {}).copy()
             self.epochs = caller_configs.epochs if caller_configs.epochs is not None else target_setup.epochs
+            target_batch_size = target_setup.data_loader.params.get("batch_size")
+            self.batch_size = caller_configs.batch_size if caller_configs.batch_size is not None else target_batch_size
+
+        self.optimizer_name = optimizer_name.lower()
+        self.criterion_name = criterion_name.lower()
 
         # Get optimizer class
         self.optimizer_class = self._get_optimizer_class(optimizer_name)
--- a/leakpro/attacks/utils/model_handler.py
+++ b/leakpro/attacks/utils/model_handler.py
@@ -35,40 +35,45 @@ class ModelHandler():
 
         caller_configs = getattr(handler.configs, caller) if caller is not None else None
         self.use_target_model_setup = caller_configs is None
+        target_setup = handler.target_model_metadata
 
         # get the bluepring for the model
         if self.use_target_model_setup:
             self.model_class = handler.target_model_blueprint.__name__
             self.model_blueprint = handler.target_model_blueprint
         else:
-            try:
-                self.model_path = caller_configs.module_path
-                self.model_class = caller_configs.model_class
-            except AttributeError as e:
-                raise ValueError("Model path or class not provided in shadow model config") from e
+            # Allow partial shadow model config by inheriting from the target setup.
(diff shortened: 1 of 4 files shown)


leakpro/schemas.py
```diff
@@ -101,13 +101,13 @@ class TargetConfig(BaseModel):
 class ShadowModelConfig(BaseModel):
     """Configuration for the Shadow models."""
 
-    model_class: Optional[str] = None
-    module_path: Optional[str] = None
+    model_class: Optional[str] = Field(default=None, description="Class name of the shadow model")
+    module_path: Optional[str] = Field(default=None, description="Path to the shadow model module")
     init_params: Optional[Dict[str, Any]] = Field(default_factory=dict, description="Model initialization parameters")
```
fazelehh: I know you have not changed this line, but: 
init_params default {} creates silent ambiguity

Root cause — [leakpro/schemas.py:106]


# current
init_params: Optional[Dict[str, Any]] = Field(default_factory=dict, ...)
Because it defaults to {} (empty dict, which is falsy), there is no way to distinguish "user didn't set this" from "user explicitly passed an empty dict".

Where it silently breaks — [leakpro/attacks/utils/model_handler.py:94-95]


# current
self.init_params = (target_setup.init_params or {}).copy()
self.init_params.update(caller_configs.init_params or {})  # {} is falsy → skipped
If a user passes shadow_model: init_params: {} intending to wipe the target's init params, the or {} short-circuits and the update is a no-op — the target's params are silently inherited anyway.

**Fix**
 two files:

[leakpro/schemas.py:106]change default to None:


init_params: Optional[Dict[str, Any]] = Field(default=None, ...)
[leakpro/attacks/utils/model_handler.py:95]check with is not None instead of truthiness:


self.init_params = (target_setup.init_params or {}).copy()
if caller_configs.init_params is not None:
    self.init_params.update(caller_configs.init_params)

leakpro/attacks/utils/shadow_model_handler.py
fazelehh: Again, not what you have changed, but isn't it better to if/raise ValueError instead of assert? it makes a more explicit error

leakpro/attacks/utils/model_handler.py
fazelehh: Again, something that you have not changed, but it is an easy fix: 

self.optimizer_config is a dict of optimizer parameters (e.g. {"lr": 0.001}). It has no 'name' key — so this error handler itself would crash with a KeyError, masking the original error entirely.

and the fix: 
raise ValueError(f"Failed to create optimizer from {self.optimizer_name}") from e

(the same thing is happening in this file line 160. I put one comment there too)

fazelehh, line 181: self.criterion_config doesn't even exist as an attribute. This would raise AttributeError instead of the intended ValueError.

fix:
raise ValueError(f"Failed to create criterion from {self.criterion_name}") from e

fazelehh (changes requested):
Some other general comments: 

1. model_class without module_path is silently confusing (model_handler.py:86–91)
If a user sets only shadow_model.model_class without shadow_model.module_path, self.model_path falls back to the target's module_path. If the target module doesn't export that class, the error message ("Failed to create model blueprint from X in Y") is clear enough, but the situation could be documented in the config schema description.

2. batch_size stored from data_loader.batch_size (shadow_model_handler.py:309)
Slightly indirect — stores data_loader.batch_size rather than self.batch_size directly. These are always equal (the dataloader was created with self.batch_size), so no bug, but self.batch_size would be more direct and avoids dependence on dataloader internals.

3. Missing test: model_class override without module_path
No test covers the case where the user provides shadow_model.model_class but not module_path. The code falls back to target's module_path. good to have a test or at least a comment.

henrikfo:
I've updated the code and broadened the PR to include the review feedback from PR #387.

1. 
> `init_params default {} creates silent ambiguity`

`ShadowModelConfig.init_params` now defaults to `None`
omitted / `None`: inherit target model init params
non-empty dict: inherit target init params, then override with shadow-specific values
explicit `{}`: clear inherited target init params

Fixes the silent issue and is a little bit broader still

2.
> `if/raise ValueError instead of assert`

The balanced-assignment runtime `assert` was replaced with `ValueError`. I also replaced similar runtime type-check asserts with explicit 'TypeError'
3.
> `It has no 'name' key`

`_get_optimizer_class()` now raises the error using the actual optimizer name
A test was added for invalid optimizer names.

4. 
> `self.criterion_config doesn't even exist`

`_get_criterion_class()` now raises the error using the actual criterion name passed

A test was added for invalid criterion names.

5. 
> `model_class without module_path is silently confusing`

The `ShadowModelConfig` schema descriptions now document that `model_class` and `module_path` fall back to the target model

6. 
> `batch_size stored from data_loader.batch_size`


Uses `batch_size=self.batch_size`  instead of reading it back from `data_loader.batch_size`

7.
> `Missing test: model_class override without module_path`

added test `leakpro/tests/mia_attacks/utils/test_shadow_model_handler.py`.

fazelehh (approved):

PR #403, opened 2026-04-01
fix: make leakpro amm compatible

LucasBeerens:
Added init file and fixed torch device bug. This enables leakpro to be installed in amm to use gia.


viktorvaladi (dismissed):
LGTM

mattiasakesson (approved):
looks correct

PR #407, opened 2026-04-16
Fix/suppress indices logger

Muhaddisabarat:
# Description

Summary of changes

* the first logger_setup function removed as the second function overrides the first. Also it works alongside add_file_handler() and handles file logging cleanly. It calls add_file_handler() later when a save path is known. The first function called both in one call.

* shadow_model_handler is modified at line 225 to add filtered meta_data (suppressing train_indices) 

# Resolved Issues
* 2 rounds of ruff checks and the last commit with clean notebook outputs.

### DIFF
--- a/examples/mia/cifar/cifar_main.ipynb
+++ b/examples/mia/cifar/cifar_main.ipynb
@@ -11,7 +11,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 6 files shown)


examples/mia/cifar/cifar_main.ipynb
```diff
@@ -292,7 +292,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": ".leakpro_dev",
+   "display_name": ".myenv (3.12.12)",
```
fazelehh, line 295: This change should not be commited

examples/mia/cifar/cifar_main.ipynb
```diff
@@ -306,7 +306,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.4"
+   "version": "3.12.12"
```
fazelehh, line 309: should not be commited

leakpro/utils/logger.py
```diff
@@ -1,9 +1,10 @@
 """Module contains the function to setup the logger for the current run."""
+
 import logging
 import os
 
 
-def setup_logger(name: str, save_file: bool=True, save_path:str=None) -> logging.Logger:
+def setup_logger(name: str, save_file: bool = True, save_path: str = None) -> logging.Logger:
```
fazelehh: why there are two setup_logger in this file?

leakpro/utils/logger.py
```diff
@@ -66,8 +67,17 @@ def setup_logger(name: str = "leakpro_log", level: int = logging.INFO) -> loggin
         console_handler.setFormatter(log_format)
         logger.addHandler(console_handler)
 
+    # Add filter to suppress train/test indices from logs
+    class SuppressIndicesFilter(logging.Filter):
+        def filter(self, record: logging.LogRecord) -> bool:
```
fazelehh: not the best practice to define the class in the function

leakpro/utils/logger.py
```diff
@@ -66,8 +67,17 @@ def setup_logger(name: str = "leakpro_log", level: int = logging.INFO) -> loggin
         console_handler.setFormatter(log_format)
         logger.addHandler(console_handler)
 
+    # Add filter to suppress train/test indices from logs
+    class SuppressIndicesFilter(logging.Filter):
```
fazelehh: Isn't it better to not have this filter here and only change line 221 in shadow_model_handler.py file to something like this: 
 logger.info(f"Metadata for shadow model {indx}: num_train={meta_data.num_train}, dataset={meta_data.dataset}")

fazelehh (changes requested):

leakpro/utils/logger.py
Muhaddisabarat: When there is two functions the second function gets over ride by the first. First one is removed.

leakpro/utils/logger.py
Muhaddisabarat: This class has been removed from here.

PR #409, opened 2026-04-24
Update cifar_handler_dpsgd.py

henrikfo:
# Description

Swapped order of train_loop(...) return to the expected "train_loss, train_acc". in "examples/mia/cifar/cifar_handler_dpsgd.py"

## Related Pull Requests

- #312

### DIFF
--- a/examples/mia/cifar/audit_dpsgd.yaml
+++ b/examples/mia/cifar/audit_dpsgd.yaml
@@ -3,8 +3,14 @@ audit:  # Configurations for auditing
   attack_list:
     - attack: lira
       online: True
-      num_shadow_models: 2
-      training_data_fraction: 0.1
+      num_shadow_models: 16
+      var_calculation: "fixed"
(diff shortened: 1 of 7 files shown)


examples/mia/cifar/audit_dpsgd.yaml
```diff
@@ -14,7 +20,7 @@ target:
   # Target model path
   module_path: "./target_model_class.py"
   model_class: "ResNet18_DPsgd" 
-  target_folder: "./target_dpsgd"
```
fazelehh, line 17: why?

fazelehh (changes requested):
Just two comments:
I think it's good to have a separate train_dpsgd_config.yaml. The reason is that I don't want these values to be overwritten later on. so if they're in a separate config file, there's less incentive to change them.
Same comment on target_folder name. I've added a comment for this in the respective file.

fazelehh (approved):

PR #415, opened 2026-04-29
LDG-5: Hashing Shadow Models

### LINKED ISSUE #346: Using old cached target logits
# Issue
## Problem Description
After retraining a target model, the code will check for any existing a cache logits. If there are logits cached already the logits from the new target model will not be cached and the old ones will be used. This might lead to unexpected erroneous results.

## Expected Behavior
The logits from the retrained target model should be used 

## What Needs to be Done
Some check that make sure that a new target model caches the new logits

Run any leakpro example notebook. Then rerun by changing a parameter in target_config.yaml (e.g. +1/-1 epochs) and then check the logs when running leakpro.run_audit(). It will say "Logits already cached at leakpro_output/attack_cache/target_logits.npy"

fazelehh:
### Problem

The caching layer did not detect when a target model was retrained, causing stale cached artifacts to be silently reused. This led to three classes of silent correctness bugs depending on what changed between runs:

- **If only the model weights changed** (same train/test split and training config, different random seed): the target logit cache was reused unchanged, so the attack was effectively evaluating the *old* model, not the new one.

- **If the training config changed** (e.g. epochs 2→4, different optimizer): shadow models were retrained with the new config but the old cached shadow models (trained with the old config) were still returned by the filter, producing a mismatch between how the target and shadow models were trained.

- **If the data split changed** (different train/test indices): shadow models trained on the old population were reused. Their stored `train_indices` referenced old data point indices that no longer matched the new audit dataset, producing a completely wrong in/out mask with no warning.

A fourth bug was also present: logit cache filenames were based on `population_hash` (now order-insensitive), but the logit array rows are positional — row `i` corresponds to position `i` in `train_indices + test_indices`. If the same samples appeared in a different order across runs, the cached logits would be silently reused with misaligned rows.

---

### Three Invalidation Signals

Three independent hashes now drive cache validity:

| Hash | Captures | Status |
|------|----------|--------|
| `target_model_hash` | Target model weights | Already existed; now used in cache filename |
| `population_hash` | Train/test index split (order-insensitive) | New — `SHA-256(sorted(train) \| sorted(test))` |
| Full training signature | Shadow model training setup (epochs, optimizer + params, criterion + params, model class, init params, online flag) | New — used in `_filter` via `_current_training_signature` / `_metadata_training_signature` |

---

### Logit Row Alignment (Option B)

`cache_logits` now saves a companion `{name}_indices.npy` file recording the exact `train_indices + test_indices` concatenation order used when computing the logit array. On load, `load_logits` compares the stored order to the current run's order:

- **Exact match** → rows returned as-is.
- **Same samples, different order** → rows reordered before returning. No recomputation.
- **Missing companion file** (old cache) → treated as a miss; both files are rewritten.
- **Different sample set** (should not occur if `_filter` is working) → treated as a miss.

---

#### ✅ Allowed — full cache reuse (no change at all)
Same target model, same data split, same training config. Everything is read from cache.

#### ✅ Allowed — shadow model reuse (same split + config, different weights)
Target model retrained with same split and config but different weight init.
- Shadow models: **reused** — same data, same config; in/out masks remain valid.
- Shadow logits: **reused or reordered** — companion file validates row alignment.
- Target logits: **recomputed** — new weights, new filename.

#### ❌ Prevented — stale shadow reuse (training config changed)
Full training signature check in `_filter` rejects mismatched shadow models.

#### ❌ Prevented — stale shadow reuse (different data split)
`population_hash` check in `_filter` rejects shadow models from a different population.

#### ❌ Prevented — stale target logits (different weights)
Cache filename includes `target_model_hash`; weight changes produce a new filename.

#### ❌ Prevented — logit row misalignment (same samples, different index order)
Companion `_indices.npy` detects order changes and reorders rows before returning.

---

### Files Changed

| File | Why |
|------|-----|
| `leakpro/utils/save_load.py` | `hash_indices` computes order-insensitive SHA-256 of the train/test split (sorts before hashing). |
| `leakpro/schemas.py` | Added `population_hash` (Optional[str]) to `ShadowModelTrainingSchema`; old metadata without the field is treated as a miss. |
| `leakpro/attacks/utils/model_handler.py` | Computes `population_hash` and `target_model_hash` at init. `cache_logits` writes a companion `_indices.npy`. New `load_logits` detects and fixes row misalignment. |
| `leakpro/attacks/utils/shadow_model_handler.py` | `_filter` uses full training signature comparison + `population_hash` check. `load_logits` delegates to base class alignment logic and handles cache misses for both shadow and target models. |
| `leakpro/attacks/mia_attacks/lira.py` | Updated `load_logits` call to use hashed target logit filename. |
| `leakpro/attacks/mia_attacks/rmia.py` | Same as above. |
| `leakpro/attacks/mia_attacks/base.py` | Same as above. |


examples/mia/cifar/cifar_main.ipynb
henrikfo, line 1: Notebook output was note cleared.

examples/mia/cifar/train_config.yaml
```diff
@@ -3,7 +3,7 @@ run: # Configurations for a specific run
   log_dir: ./target # String for indicating where to save all the information, including models and computed signals. We can reuse the models saved in the same log_dir.
 
 train:
-  epochs: 25
```
henrikfo: Should not change default/out-of-the-box parameters (Will be changed back though by other PRs anyways?)

examples/mia/cifar/audit.yaml
```diff
@@ -1,11 +1,11 @@
 audit:  # Configurations for auditing
   random_seed: 1236  # Integer specifying the random seed
   attack_list:
-# # RMIA
-#     - attack: rmia
-#       online: False
-#       num_shadow_models: 2
```
henrikfo: I think RMIA or LiRA is the better alternative, rather then SeqMIA as the default attack. But maybe a higher value on num_shadow_models? Like 5 / 8  as the default?

leakpro/attacks/mia_attacks/base.py
henrikfo, line 155: This is great! Although it's done for base, rmia and lira but what about other attacks?

fazelehh: Only base.py, rmia.py, and lira.py call load_logits with a target model name. All three have been updated. No other MIA attacks use this call directly.

examples/mia/cifar/audit.yaml
fazelehh: done

examples/mia/cifar/train_config.yaml
fazelehh: done

examples/mia/cifar/cifar_main.ipynb
fazelehh: done

leakpro/attacks/utils/shadow_model_handler.py
```diff
@@ -404,7 +417,7 @@ def load_logits(self:Self, name:str=None, indx:int=None) -> np.ndarray:
         """Load model logits."""
 
         if name is None and indx is not None:
-            name = f"sm_{indx}"
+            name = f"sm_{indx}_{self.population_hash}"
```
henrikfo, line 409: population_hash is fine for reusing the same shadow model, but it is also used for the logits cache. Logits are just a positional array, so the row order must match the current `train_indices + test_indices` order. If the same train/test sets are ever saved in a different order, this can reuse a logits file with misaligned rows. Could we store/validate the row indices with the logits, or document/enforce that metadata order is stable?
If the ordering would somehow be messed up, the logits themselves become stale or misaligned with the labels/membership.
I'm thinking if this is a scenario could be an edge case somehow? What do you think?

henrikfo (dismissed):

leakpro/attacks/utils/shadow_model_handler.py
fazelehh: you're right that using a sorted (order-insensitive) hash for the logit cache creates a risk: two runs with the same set of indices in a different order would produce the same hash, but the cached logit rows would be misaligned with the current train_indices + test_indices ordering.

Fix applied: changed hash_indices to be order-sensitive by removing the np.sort before hashing. The bytes fed to SHA-256 now reflect the actual order of train_indices and test_indices, so any change in ordering produces a new hash, a cache miss, and a fresh recomputation. This guarantees that a loaded logit file always has rows aligned with the current index order.

The trade-off is that shadow models would be retrained if the same indices appear in a different order between runs, but this is a safe failure (wasted compute, never wrong results), and in practice the index order from the data split is always deterministic given the same random seed.

fazelehh: I think this part is not fully correct "For shadow models, we only need to know: “is this the same train/test population?” For that, order should not matter. If the same samples are listed in another order, we should still be able to reuse the same shadow model." 
We also need to see if the learning parameters are the same or not. like if we are using the blueprint of a target model. Like if the lr in the target model has changed then the shadow model need to be changed too.

henrikfo:
I think there are two slightly different things these hashes can be used for.

For shadow models, we only need to know: “is this the same train/test population?” For that, order should not matter. If the same samples are listed in another order, we should still be able to reuse the same shadow model. 
For the logits, the logits file is just rows in an array. Row 0 is assumed to belong to the first sample in `train_indices + test_indices`, row 1 to the second sample, and so on. So for logits, order does matter.

That means one hash cannot cleanly serve both purposes unless we accept either wrong row alignment or unnecessary shadow retraining. A solution could be to keep the population_hash order insensitive to use for the shadow model reuse. And a separate order sensitive hash for logits cache files, or store the row indices with the logits and validate/reorder when loading

Also, the current code still sorts inside hash_indices(...), but I saw that some commits ago there was np.array() instead of np.sort(), maybe they accidentally got reverted?

Another thing I noticed, was there a reason for dropping online from "_filter(..., online=online)" ? I think this can also affect the hash, atleast the signature fucntion

leakpro/attacks/utils/shadow_model_handler.py
fazelehh: So there are two layers of reusing shadow models: reusing trained shadow models, and reusing the logits of that shadow models.
There are three options that I can see: 


 ##Option A: Two separate hashes

population_hash (sorted, order-insensitive) → used in _filter for shadow model reuse
logit_hash (unsorted, order-sensitive) → used for logit cache filenames
Simple to implement. But you now have two hash fields in the metadata schema and two concepts to keep track of.

##Option B: Store indices alongside logits

Keep population_hash (sorted) for shadow model reuse
When saving logits, also save the train_indices + test_indices array alongside the .npy file
When loading logits, check if the stored order matches the current run — if not, reorder the rows before returning. More robust. If indices are in a different order, you can silently fix alignment instead of failing or misaligning. But loading logic gets more complex.

##Option C: Don't cache logits by population order — always recompute from the shadow model


I vote for option B. What do you think?

fazelehh:
So there are two layers of reusing shadow models: reusing trained shadow models, and reusing the logits of that shadow models.
There are three options that I can see:

##Option A: Two separate hashes

population_hash (sorted, order-insensitive) → used in _filter for shadow model reuse
logit_hash (unsorted, order-sensitive) → used for logit cache filenames
Simple to implement. But you now have two hash fields in the metadata schema and two concepts to keep track of.

##Option B: Store indices alongside logits

Keep population_hash (sorted) for shadow model reuse
When saving logits, also save the train_indices + test_indices array alongside the .npy file
When loading logits, check if the stored order matches the current run — if not, reorder the rows before returning. More robust. If indices are in a different order, you can silently fix alignment instead of failing or misaligning. But loading logic gets more complex.

##Option C: Don't cache logits by population order — always recompute from the shadow model

I vote for option B. What do you think?

leakpro/attacks/utils/shadow_model_handler.py
henrikfo: > I think this part is not fully correct "For shadow models, we only need to know: “is this the same train/test population?” For that, order should not matter. If the same samples are listed in another order, we should still be able to reuse the same shadow model." We also need to see if the learning parameters are the same or not. like if we are using the blueprint of a target model. Like if the lr in the target model has changed then the shadow model need to be changed too.

No you're right! But this should be covered by _current_training_signature(..) I think already

henrikfo:
> So there are two layers of reusing shadow models: reusing trained shadow models, and reusing the logits of that shadow models. There are three options that I can see:
> 
> ##Option A: Two separate hashes
> 
> population_hash (sorted, order-insensitive) → used in _filter for shadow model reuse logit_hash (unsorted, order-sensitive) → used for logit cache filenames Simple to implement. But you now have two hash fields in the metadata schema and two concepts to keep track of.
> 
> ##Option B: Store indices alongside logits
> 
> Keep population_hash (sorted) for shadow model reuse When saving logits, also save the train_indices + test_indices array alongside the .npy file When loading logits, check if the stored order matches the current run — if not, reorder the rows before returning. More robust. If indices are in a different order, you can silently fix alignment instead of failing or misaligning. But loading logic gets more complex.
> 
> ##Option C: Don't cache logits by population order — always recompute from the shadow model
> 
> I vote for option B. What do you think?

I think option B is the better solution too! Let's do that

henrikfo (approved):

PR #417, opened 2026-04-30
Broken links

fazelehh:
## Fix broken links + add use-case landing pages

**What changed:**
- Fixed 4 broken links in the README "Real world examples" table — the target `.md` files did not exist
- Created use-case landing pages under `examples/use-cases/` (one per use case), each describing the scenario and linking to the relevant public example folders per attack type
- Added 3 new research publications to the Research Outputs section (Lassila et al. NeurIPS 2025, Valadi et al. arXiv 2025, Johansson et al. arXiv 2025)


TheColdIce (commented):
Nice!

TheColdIce (approved):

PR #418, opened 2026-05-04
Splitting handlers

fazelehh:
## Summary

This PR separates the monolithic `CifarInputHandler` into three focused files, and extends the LeakPro core to accept a dedicated model handler alongside the data handler.

### Core changes (`leakpro/`)

- **`abstract_input_handler.py`**: Added `role` parameter to `__init_subclass__` (`"full"` / `"data"` / `"model"`), allowing handler subclasses that provide only `UserDataset` or only `train()`/`eval()` without triggering a validation error.
- **`leakpro.py`**: `LeakPro.__init__` now accepts an optional `model_handler` argument. When provided, `UserDataset` comes from the first handler and `train()`/`eval()` come from the model handler. Fully backward compatible — existing single-handler usage is unchanged.
- **`mia_handler.py`**: Updated `MIAHandler.__init__` to accept a separate `training_handler`, pulling `UserDataset` from the data handler and `train()`/`eval()` from the training handler.

### Example changes (`examples/mia/cifar/`)

- **`cifar_data_handler.py`** *(new)*: `CifarDataHandler` — `UserDataset` only (normalization, augmentation).
- **`cifar_model_handler.py`** *(new)*: `CifarModelHandler` — standard `train()`/`eval()` with val split, early stopping, and LR scheduler.
- **`cifar_model_handler_dpsgd.py`** *(new)*: `CifarModelHandlerDPsgd` — DP-SGD `train()`/`eval()` using Opacus.
- **`cifar_handler.py`** *(removed)*: Replaced by the three files above.
- **`cifar_main.ipynb`**, **`cifar_dpsgd.ipynb`**, **`ramia_satml.py`**: Updated to use the new handlers and the two-argument `LeakPro()` API.


leakpro/leakpro.py
```diff
@@ -73,21 +79,26 @@ def __init__(self:Self, user_input_handler:AbstractInputHandler, configs_path:st
...
-            user_input_handler (AbstractInputHandler): The user-defined input handler
-            configs (dict): The configuration object
+            user_input_handler (AbstractInputHandler): Data handler providing UserDataset.
+            configs (dict): The configuration object.
+            model_handler (AbstractInputHandler, optional): Model handler providing train/eval.
+                Falls back to user_input_handler when None.
 
         """
+        # Resolve which handler provides train/eval
+        training_handler = model_handler if model_handler is not None else user_input_handler
```
henrikfo, line 87: Since the training handler is now split and "user_input_handler" will handle the data, is it necessary/use full with this fallback? can we instead make a verification at this point that the model_handler is not None and contains a train(...) function etc.

henrikfo: > Since the training handler is now split and "user_input_handler" will handle the data, is it necessary/use full with this fallback? can we instead make a verification at this point that the model_handler is not None and contains a train(...) function etc.

 Or if we want to support the older style aswell, we need to make sure that if no model handler is given, the user_input_handler has a train() etc. methods still.  Shadow models will otherwise return "None" from `self.handler.train(...)` since it will inherit, through the user_input_handler, the train() method from the abstract_input_handler()

examples/mia/cifar/cifar_model_handler.py
```diff
@@ -0,0 +1,131 @@
...
+        epochs: int = None,
+    ) -> TrainingOutput:
+        val_split = 0.1
+        patience = 10
+        dataset = dataloader.dataset
+        val_size = int(len(dataset) * val_split)
+        train_size = len(dataset) - val_size
+        train_subset, val_subset = torch.utils.data.random_split(dataset, [train_size, val_size])
+
+        if hasattr(dataset, "augment"):
```
henrikfo, line 39: I think this was already an issue in the old `cifar_handler.py`, so it is not really anything introduced here by the splitting handlesr PR, but since the training loop is moved here it would be good to fix it.

`random_split` returns two `Subset`s that point to the same underlying dataset object. So when we do `val_subset.dataset.augment = None`, we also disable augmentation for `train_subset.dataset`.

That means if someone creates the CIFAR dataset with, e.g. `augment_strength="easy"`, the code will silently train without augmentation even though the dataset was configured to use it. Could we instead create a separate validation dataset instance with the same normalization params but no augmentation, so train and validation do not share mutable augmentation state?

examples/mia/celebA_HQ/celebA_hq_dpsgd.ipynb
```diff
@@ -106,54 +106,7 @@
    "execution_count": null,
```
henrikfo, line 104: <p>We can remove this "==1.5.3" aswell.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aidotse/LeakPro/pull/418/discussion/#comment-3218854071'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='2'/>

henrikfo, line 104: <pre class="ql-syntax" spellcheck="false">Line #3.    train_acc = [acc*100 for acc in train_result.metrics.extra["accuracy_history"]]
</pre><p>This "100*" should be removed, but I can do it for my celebA training patch.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aidotse/LeakPro/pull/418/discussion/#comment-3218854076'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='11'/>

examples/mia/cifar/cifar_dpsgd.ipynb
```diff
@@ -2,15 +2,17 @@
  "cells": [
```
henrikfo, line 2: <p>Do we want to remove this instruction text?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aidotse/LeakPro/pull/418/discussion/#comment-3218854082'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='8'/>

henrikfo, line 2: <pre class="ql-syntax" spellcheck="false">Line #9.        'epochs': 20,
</pre><p>Should still be epochs=epochs, target_epsilon=30 and  max_grad_norm=1.2</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/aidotse/LeakPro/pull/418/discussion/#comment-3218854087'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='9'/>

henrikfo (changes requested):
There was two issues that I could find, of which is directly related to the PR (splitting handlers) if this is adressed we can merge! There was some other minor stuff too. 

I found another issue too, but its not directly related to the PR. Making this into an separate issue or solving it, either of them is fine for me for merging

henrikfo (dismissed):
Looks great, good job! Just resolve the conflicts and we are good to go! If there are any changes hapening by resolving those conflicts you feel like is large enough to have someone review it, ping me again or ask for a re-review! 

Also, could you just verify if this is a problem with ReviewNB or if you see if on your side aswell? 

https://app.reviewnb.com/aidotse/LeakPro/pull/418/?path=c46a76a4&cell_index=9&sha=cddcf76875aa48187ce45362793550e48fe154e2

In ReviewNB it looks like just one straight line, one character wide for 513 lines. But I'm checking out the branch on a my machine, then it seems fine.

henrikfo (approved):
Good job!

PR #420, opened 2026-05-07
Remove unused example

### LINKED ISSUE #419: Old copy files in gia example
There is old files in examples/gia/bn_research: 
- main_copy.py
- main_copy_copy.py

Can these be removed or better named?

viktorvaladi:
Removes example that should be removed


TheColdIce (approved):
Great work Viktor! :D

PR #424, opened 2026-05-19
LDG-8: Patch dpsgd config

henrikfo:
# Description

Summary of changes
Changed the parameters used for DP in the CelebA example

## Resolved Issues
Make the examples work out-of-the-box.


examples/mia/celebA_HQ/celebA_hq_dpsgd_handler.py
```diff
@@ -0,0 +1,440 @@
+"""Input handler for CelebA-HQ experiments with optional DP-SGD training."""
```
fazelehh: This file should be fixed in terms of diff. It should only show few lines that were changed in this PR.

fazelehh (changes requested):

TheColdIce (dismissed):
## Fix this then ok to merge
- [ ] add `target_dpsgd/*` and `leakpro_dpsgd_output/*` to the .gitignore in the celeba and cifar examples.

fazelehh (approved):

PR #427, opened 2026-05-26
LDG-19: adding online to filter

fazelehh:
## Problem

The `online` flag (online vs. offline shadow model attack mode) was being passed to `ShadowModelHandler.create_shadow_models()` and stored as part of the shadow model cache key. This meant that shadow models trained for an online attack and shadow models trained for an offline attack were treated as incompatible — even though **the training procedure is identical in both cases**. Switching between modes would trigger unnecessary retraining of all shadow models.

The flag was also stored in `ShadowModelTrainingSchema` and threaded through `_filter()`, `_current_training_signature()`, and `_metadata_training_signature()`, adding false complexity to the caching logic.

Additionally, `rmia.py` had `online=True` hardcoded in its `_prepare_shadow_models()` call, meaning shadow model metadata would be tagged as online even when the attack was configured as offline. The `_run_attack` log message also always said "online" regardless of the configured mode.

## Solution

- Removed `online` from `ShadowModelHandler.create_shadow_models()`, `_filter()`, `_current_training_signature()`, and `_metadata_training_signature()`
- Removed the `online` field from `ShadowModelTrainingSchema`
- Removed `online=self.online` / `online=True` call-site arguments across all MIA attack files (`base.py`, `dts.py`, `lira.py`, `multi_signal_lira.py`, `oslo.py`, `rmia.py`, `yoqo.py`)
- Fixed RMIA's hardcoded log message to correctly reflect online vs. offline mode

The `online` config option still exists on each attack class and continues to control scoring behaviour — which shadow model outputs to average over, whether to compute `out_indices`, and whether to apply the offline scale factor. That is the only place where online vs. offline actually matters.

## How Has This Been Tested?

- Existing shadow model handler unit tests updated to reflect the removed parameter
- No change to scoring logic

### DIFF
--- a/leakpro/attacks/utils/shadow_model_handler.py
+++ b/leakpro/attacks/utils/shadow_model_handler.py
@@ -91,14 +91,13 @@ class ShadowModelHandler(ModelHandler):
             return tuple(self._freeze_value(item) for item in value)
         return value
 
-    def _current_training_signature(self:Self, data_size:int, online:bool) -> tuple:
+    def _current_training_signature(self:Self, data_size:int) -> tuple:
         """Build the effective training signature for the current shadow setup.
 
         Args:
         ----
             data_size (int): Number of samples that will be used to train a
                 shadow model.
-            online (bool): Whether the shadow models are created in online mode.
 
         Returns:
         -------
@@ -118,7 +117,6 @@ class ShadowModelHandler(ModelHandler):
             self._freeze_value(self.loss_config),
             self.epochs,
             self.batch_size,
-            online,
             self._freeze_value(self.init_params),
         )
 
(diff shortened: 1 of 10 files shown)


leakpro/attacks/utils/shadow_model_handler.py
```diff
@@ -251,7 +251,7 @@ def create_shadow_models(
 
         # Get the size of the dataset
         data_size = int(len(shadow_population)*training_fraction)
-        all_indices, filtered_indices = self._filter(data_size)
+        all_indices, filtered_indices = self._filter(data_size, online)
```
henrikfo: I think online=online is the correct function call here!

henrikfo (changes requested):

henrikfo:
Flag for target_model_hash sensitivity.

henrikfo (approved):

henrikfo:
> Flag for target_model_hash sensitivity.

Will be handled in a separate PR

PR #431, opened 2026-06-18
Feature/ldg 28 gdd ens mia example

TheColdIce:
# Description

Adds a new MIA example under `examples/mia/gdd_ens` that audits a tumor-type classifier in the style of **GDD-ENS** ([Darmofal et al., *Cancer Discovery* 2024](https://doi.org/10.1158/2159-8290.CD-23-0996); [code](https://github.com/mmdarmofal/GDD_ENS)) for membership leakage via the split-handler API. It ships three variants: a single-MLP baseline, a DP-SGD trade-off sweep, and a faithful 10-MLP ensemble that audits a real-model-style target.

* **Baseline example** (`gdd_main.ipynb`): builds a population from the GDD full feature table (one sample per patient, 38 label-encoded tumor types, GDD preprocessing up to but not including its train/test split + oversampling), trains a fresh single-MLP target, and runs the audit with **LiRA + RMIA + population**. Handlers: `GddDataHandler` (data) + `GddModelHandler` (model); target arch in `utils/gdd_model.py`, one-time CSV→tensor prep in `utils/gdd_data.py`.
* **DP-SGD variant** (`gdd_dpsgd.ipynb`): trains the same single-MLP target privately with Opacus across an epsilon grid and re-runs the audit at each budget, for a utility (test accuracy) vs. privacy (ROC AUC / TPR @ low FPR) trade-off, with the non-DP baseline as the epsilon = infinity point. Shadows are trained under DP too — `GddMLP_DPsgd` records `dpsgd=True` in the target metadata, so shadows rebuilt from it inherit the same guarantee and the LiRA/RMIA likelihood ratio is calibrated against DP-trained shadows. Each epsilon gets its own `output_dir` because LeakPro's shadow-model cache key does not include the DP noise level (reusing one dir would silently reuse the first epsilon's shadows). Files: `gdd_model_handler_dpsgd.py`, `utils/gdd_model_dpsgd.py`, `audit_dpsgd.yaml`, `train_config_dpsgd.yaml`.
* **Ensemble variant (real-model-style)** (`gdd_ensemble_main.ipynb`): audits a faithful **10-MLP GDD-ENS ensemble** (member architectures from the paper's Supplementary Table S6) instead of the single-MLP proxy. `forward` returns `log(mean_k softmax(member_logits_k))`, the paper's softmax-averaging, so LeakPro's signal softmax recovers the true ensemble probability. The member/non-member split **reproduces the paper's deterministic `split_data.py`** (80/20, `random_state=0`, test de-duplicated against the train patients), so the membership ground truth is the real one without needing Supplementary Table S1. **One shadow = one ensemble**: `shadow_model` is blank, so each shadow is itself a full `GddEnsemble` rebuilt from the target metadata and trained by `GddEnsembleModelHandler` (total trainings = `10 x (1 + num_shadow_models) x epochs` MLPs). Runs **LiRA + RMIA only** (see below). Files: `gdd_ensemble_handler.py`, `utils/gdd_ensemble.py`, `utils/gdd_paper_data.py`, `audit_ensemble.yaml`, `train_config_ensemble.yaml`.

**Scope / caveats (documented in the README):**
- The baseline/DP-SGD variants audit a **proxy, not the published GDD-ENS model** — a single, lightly-trained MLP (20 epochs, no upsampling, raw unscaled features) rather than the 10-MLP HPO ensemble. They are a **methodological demonstration**, not a privacy assessment of the deployed model. The ensemble variant reproduces the architecture and the real training split, but still deviates deliberately (no oversampling, no per-member stratified folds, no per-member HPO re-run) so target and shadows train identically for clean LiRA/RMIA calibration; see the README's "Ensemble" section.
- **No population (P) attack in the ensemble audit, by design.** The paper-faithful split partitions the *entire* population into members (80%) + non-members (20%), so `audit_size == population_size`. The population attack fits its loss thresholds on data outside both train and test, i.e. it needs a held-out reference partition — with none left, `AttackP` raises *"the audit dataset is the same size as the population dataset"* and LeakPro skips it. The baseline avoids this because it carves only `f_train` + `f_test` fractions and leaves the remainder in the population. LiRA/RMIA are unaffected (their reference comes from shadow models). The omission and rationale are documented in `audit_ensemble.yaml`, the README, and the notebook intro.
- The example does **not** ship data: GENIE is controlled-access (~860 MB feature table). The README documents how to obtain `msk_solid_heme_ft.csv`. Generated configs, sweep outputs, target dirs, and the feature table are gitignored.

## Resolved Issues

- Implements LDG-28 (create GDD_ENS MIA example).

## How Has This Been Tested?

- Baseline audit run end-to-end on the GDD full feature table; on a representative run the shadow-based attacks beat the population baseline (RMIA ≈ 0.68, LiRA ≈ 0.64 ROC AUC) while population sits at ~0.5 (no shadow models → no membership signal, the expected floor).
- DP-SGD sweep runs end-to-end across the epsilon grid and produces a trade-off CSV. Note: with this lightly-trained single-MLP proxy on 38 imbalanced classes, DP training drives utility down toward the majority-class floor, so the current sweep is best read as a wiring/methodology demonstration rather than an informative trade-off curve.
- Ensemble audit run end-to-end as a reduced smoke run (2 epochs, 2 shadow ensembles): LiRA and RMIA complete and produce a report; the population attack is correctly skipped with the "audit dataset is the same size as the population dataset" message, confirming the by-design behaviour above. A full paper-faithful run (8 shadows x 200 epochs) is a long GPU job and has not been run end-to-end here.

## Related Pull Requests

- _none_


examples/mia/gdd_ens/utils/gdd_ensemble.py
```diff
@@ -0,0 +1,102 @@
...
+import torch
+from torch import Tensor, nn
+from torch.nn import functional as F  # noqa: N812
+
+# Per-member hyperparameters from GDD-ENS Supplementary Table S6 (all 10 MLPs). Architecture fields
+# (num_fc_layers, num_fc_units, dropout_rate) define each member; learning_rate / weight_decay are
+# read by the model handler to build per-member optimizers during training.
+TABLE_S6 = (
+    {"num_fc_layers": 1, "num_fc_units": 1376, "dropout_rate": 0.5,      "learning_rate": 1.46e-4, "weight_decay": 7.87e-4},
+    {"num_fc_layers": 2, "num_fc_units": 1051, "dropout_rate": 0.5,      "learning_rate": 2.22e-4, "weight_decay": 4.8e-5},
```
fazelehh: **Member 2 width contradicts the released GDD-ENS weights.** I loaded the authors' `data/ensemble_models/fold_2.pt`; its linear layers are `Linear(4487→950)`, `Linear(950→950)`, `Linear(950→950)`, `Linear(950→38)` — i.e. `num_fc_units = 950`, not 1051 (the layer count of 2 is correct). All 9 other members match this table exactly; only member 2 differs. For a reproduction of the deployed model this should be **950**.

Since one of the only four weight-verifiable numbers was off, worth re-checking the rest of the table against the actual Supplementary Table S6 — `dropout_rate` / `learning_rate` / `weight_decay` can't be recovered from the saved weights, so they're currently unverified.

examples/mia/gdd_ens/README.md
```diff
@@ -0,0 +1,187 @@
...
+- **Optimizer matches the baseline (Adam)** so the only thing changing across the sweep is the
+  added noise + per-sample clipping; δ defaults to `1e-5` (< 1/N). Tune ε, δ, `max_grad_norm`,
+  and `num_shadow_models` in the configs. On 38 imbalanced classes, small ε crushes rare-class
+  utility first — report macro accuracy, not just overall.
+- Requires `opacus` (the LeakPro environment ships it; `pip install opacus` otherwise) and a GPU
+  for a full sweep.
+
+### Ensemble (real-model-style audit)
+
+`gdd_ensemble_main.ipynb` audits a **faithful 10-MLP GDD-ENS ensemble** instead of the single-MLP
```
fazelehh: **"faithful" overstates the ensemble.** The architecture and the softmax-averaging rule are faithful, but the *training* is not: the paper decorrelates its 10 members via per-member StratifiedShuffleSplit folds + independent per-member HPO + oversampling, whereas here all 10 members train on the **identical in-split**, differing only by architecture and random init/dropout. That materially changes ensemble diversity — and therefore the membership-leakage profile being audited — so the MIA numbers here shouldn't be read as GDD-ENS's actual leakage. The deviations are documented further down, but suggest softening the headline to e.g. "architecture-faithful" and stating the training deviation up front.

examples/mia/gdd_ens/audit_ensemble.yaml
```diff
@@ -0,0 +1,40 @@
...
+audit:
+  random_seed: 1234
+  attack_list:
+    - attack: lira
+      online: False
+      num_shadow_models: 8
+      training_data_fraction: 0.5
+      var_calculation: "fixed"
+    - attack: rmia
+      online: True
```
fazelehh, line 18: **Minor — RMIA temperature acts on log-probs for the ensemble.** RMIA pulls raw `ModelLogits` then applies `softmax_logits(·, temperature)`. Because `GddEnsemble.forward` returns `log(mean softmax)`, RMIA effectively computes `softmax(log p / T)` — a *tempered* ensemble probability, not `p` exactly (unlike LiRA's `get_rescaled_logits`, whose stable softmax recovers `p` exactly). It's self-consistent since target and shadows are both `GddEnsemble`, so not a calibration bug, but the semantics differ from the single-MLP case and from LiRA. Worth a one-line note.

examples/mia/gdd_ens/train_config_dpsgd.yaml
```diff
@@ -0,0 +1,36 @@
...
+train:
+  epochs: 20
+  batch_size: 32
+  virtual_batch_size: 32     # max physical batch for Opacus BatchMemoryManager (<= batch_size)
+  optimizer: Adam            # DP-Adam: Adam + per-sample clipping + noise (matches baseline)
+  learning_rate: 0.0001
+  weight_decay: 0.001
+
+dp:
+  target_delta: 1.0e-5       # < 1/N (N ~ 36,976) per the usual delta << 1/N guideline
```
fazelehh: **Minor — wrong N in the delta comment.** DP-SGD trains on `f_train=0.3` of the *de-duplicated* (one-per-patient) population, not the full ~36,976 `train`-category rows. The `delta << 1/N` guideline still holds, but the cited `N ~ 36,976` refers to a different set than the one actually trained on; suggest updating the comment to the real training-set size.

examples/mia/gdd_ens/utils/gdd_paper_data.py
```diff
@@ -0,0 +1,136 @@
...
+        nunique_train = train_df[feature_cols].nunique()
+        feature_cols = [c for c in feature_cols if nunique_train[c] > 1]
+
+    members_X = train_df[feature_cols].to_numpy(dtype=np.float32)
+    nonmembers_X = test_df[feature_cols].to_numpy(dtype=np.float32)
+    features = np.concatenate([members_X, nonmembers_X], axis=0)
+
+    # Encode targets over the combined population so every label seen anywhere is mapped.
+    encoder = LabelEncoder()
+    all_labels = pd.concat([train_labels, test_labels]).to_numpy()
```
fazelehh, line 113: **Minor — encoder fit over members + non-members.** Fitting the `LabelEncoder` on `concat(train_labels, test_labels)` means a tumor type that happens to appear only among non-members gets an output neuron the target never trains on. Unlikely to bite with 38 well-populated classes, but worth a guard or a comment noting the assumption.

fazelehh (commented):
Code review focused on faithfulness to the GDD-ENS paper + the authors' released code. The LeakPro integration is correct (verified the `log(mean softmax)` forward recovers the true ensemble probability through LeakPro's stable softmax, and the split reproduction matches `split_data.py`). Inline: one confirmed bug verified against the authors' released weights, one framing point on "faithful", and the four minor notes from the review.

examples/mia/gdd_ens/gdd_dpsgd.ipynb
```diff
@@ -0,0 +1,223 @@
...
...on)\n\n    target_model.to('cpu')\n    with open(os.path.join(log_dir, 'target_model.pkl'), 'wb') as f:\n        save(target_model.state_dict(), f)\n    meta_data = LeakPro.make_mia_metadata(\n        train_result=train_result, optimizer=optimizer, loss_fn=criterion,\n        dataloader=train_loader, test_result=test_result, epochs=epochs,\n        train_indices=train_indices, test_indices=test_indices, dataset_name=dataset_name)\n    with open(os.path.join(log_dir, 'model_metadata.pkl'), 'wb') as f:\n        pickle.dump(meta_data, f)\n    print(f'  target train acc {train_result.metrics.accuracy:.4f} | '\n          f'test acc {test_result.accuracy:.4f}')\n\n    audit_path = write_audit_config(eps)\n    leakpro = LeakPro(GddDataHandler, audit_path, model_handler=GddModelHandlerDPsgd)\n    mia_results = leakpro.run_audit(create_pdf=True)\n\n    for res in mia_results:\n        rows.append({\n            'epsilon': eps,\n            'test_acc': test_result.accuracy,\n            'attack': attack_name(res),\n            'roc_auc': float(getattr(res, 'roc_auc', np.nan)),\n            'tpr@fpr=1e-2': tpr_at_fpr(res, 1e-2),\n            'tpr@fpr=1e-3': tpr_at_fpr(res, 1e-3),\n        })"
```
fazelehh: **Minor — fragile result extraction.** `getattr(res, 'roc_auc', np.nan)` silently yields NaN (and the row still survives into the CSV/plots) if that attribute is ever named differently across LeakPro versions; likewise `tpr_at_fpr` interpolates without asserting the ROC arrays are present. Cosmetic, but a missing-attribute assertion would fail loudly instead of producing silent NaNs in the trade-off table.

examples/mia/gdd_ens/utils/gdd_ensemble.py
TheColdIce: Good catch — confirmed and fixed in 0f5bd34. Set member 2 to `num_fc_units: 950` to match the released `fold_2.pt`. Left the other 9 members as-is (their layer counts + widths already match the weights) and added a comment noting that `dropout_rate`/`learning_rate`/`weight_decay` can't be recovered from the saved weights, so they stay as printed in Table S6 (unverified).

examples/mia/gdd_ens/README.md
TheColdIce: Agreed. In 0f5bd34 I softened the headline to "architecture-faithful" (README + the `gdd_ensemble.py` module docstring) and added an up-front callout that the *training* deviates (identical in-split, no per-member folds/HPO/oversampling), so the MIA numbers must not be read as the deployed GDD-ENS's leakage. The full deviation list stays below it.

examples/mia/gdd_ens/audit_ensemble.yaml
TheColdIce: Right — added the one-line note in 0f5bd34 next to the `rmia` entry in `audit_ensemble.yaml`, spelling out that RMIA computes `softmax(log p / T)` (a tempered ensemble probability) whereas LiRA's `get_rescaled_logits` recovers `p`, and that it's self-consistent since target and shadows are both `GddEnsemble`.

examples/mia/gdd_ens/train_config_dpsgd.yaml
TheColdIce: Fixed in 0f5bd34. Updated the comment to state that N is the DP *training* set = `f_train` (0.3) x the de-duplicated one-per-patient population (~36,976), i.e. ~11k rows, not the full population — `delta << 1/N` still holds comfortably at 1e-5.

examples/mia/gdd_ens/utils/gdd_paper_data.py
TheColdIce: Added a guard in 0f5bd34: we now `warnings.warn` if any tumor type appears only among non-members (a class neuron the target never trains on), rather than silently encoding it. Kept the fit over the combined population so the label range is stable; the warning makes the assumption fail loudly if it's ever violated.

examples/mia/gdd_ens/gdd_dpsgd.ipynb
TheColdIce: Fixed in 0f5bd34. Replaced `getattr(res, 'roc_auc', np.nan)` with a direct `res.roc_auc` read (raises `AttributeError` on a rename instead of emitting a silent NaN row), and added an assertion in `tpr_at_fpr` that the ROC arrays are present before interpolating.

examples/mia/gdd_ens/utils/gdd_ensemble.py
TheColdIce: Rechecked the full table against Supplementary Table S6: all 10 members' `num_fc_layers`, `dropout_rate`, `learning_rate`, and `weight_decay` match the paper exactly. The only discrepancy is member 2's `num_fc_units` — the **paper prints 1051 but the released `fold_2.pt` is 950**, so this is a genuine paper-vs-weights mismatch, not a transcription error. We keep 950 to reproduce the deployed model. Recorded the full-table verification in the `TABLE_S6` comment (d7e60b6).

examples/mia/gdd_ens/gdd_ensemble_main.ipynb
```diff
@@ -0,0 +1,159 @@
...
...MB) and caches the prepared population.\n\n**Deviations from the paper (documented in the README):** no oversampling and no per-member\nstratified folds (target and shadows are trained identically for clean LiRA/RMIA calibration). The\ntarget is always trained from scratch; the released GDD-ENS weights are not used (they ship as a\nfragile torch-1.4 whole-model pickle).\n\n**Why no population attack here.** The single-MLP baseline (`gdd_main.ipynb`) audits LiRA / RMIA / **population**, but this ensemble audit runs only **LiRA + RMIA**. The population (P) attack fits its loss thresholds on data outside both the train and test sets, so it needs a held-out reference partition. The paper-faithful split partitions the *entire* population into members (80%) + non-members (20%), leaving `audit_size == population_size` and nothing to threshold on — `AttackP` raises *\"the audit dataset is the same size as the population dataset\"* and is skipped. The baseline avoids this by carving only `f_train` + `f_test` fractions and leaving the rest in the population. LiRA and RMIA are unaffected: their reference comes from shadow models, not a held-out population slice. See the README for details.\n"
```
fazelehh: **Minor — notebook intro still says "faithful," not "architecture-faithful."** The README and `gdd_ensemble.py` docstring were reworded per the comment above, but this notebook's own intro — the first thing a reader opens — still reads *"Membership-inference attack against a **faithful 10-MLP GDD-ENS ensemble**"*, the same overstatement. Carry the same reframing + training-deviation callout here for consistency.

fazelehh (dismissed):

fazelehh (approved):

PR #423, opened 2026-05-13
LDG-4: Web app LoS

fazelehh:
# Description

Adds tabular / Length-of-Stay (MIMIC-III) support to the LeakPro webapp.

* **Binary classification support in the train step**: 2-class datasets are detected automatically and trained with a single-logit head + `BCEWithLogitsLoss` (multi-class keeps `CrossEntropyLoss`). Custom archs that output 2 logits still work — binary-ness is derived by probing the actual model output.
* **Tabular MLP preset** in Step 3 with auto-detected `num_features` / `num_classes`.
* **Audit support for binary/tabular models**: the default shadow-model handler handles the BCE criterion, `data_modality` follows the job's data type (was hardcoded to `image`), and the signal extractor handles single-logit heads in `get_loss` (needed by loss_traj).
* **Tabular results view**: clicking a record in the top-risk table opens a modal with the row's feature values (new `/jobs/{id}/sample_data/{index}` endpoint), replacing the image thumbnail column.
* **Frontend build fix**: resolved two pre-existing `tsc` errors (`DataMeta.label_column`, Histograms plotly typing) so `npm run build` passes.

Note: the LoS-specific upload helpers and datasets (`webapp_los_files/`) are intentionally **not** part of this PR — they are user-uploaded webapp inputs, handled the same way as `webapp_celebA_files` / `webapp_cifar_files` (untracked + gitignored).

## How Has This Been Tested?

* Toy binary tabular data through `_default_train`/`_default_eval`, the worker's default handler (binary + multi-class regression check), and `get_loss`/`get_rescaled_logits` for both 1-logit and CrossEntropy models — all pass.
* Real LoS data (`data_LR.pkl`) through the webapp train path: binary detection → 1-logit LR → BCE, 1 epoch, test acc 0.64.
* `npm run build` (tsc + vite) passes.

The RMIA/BASE/loss_traj crashes reported in review are root-caused and fixed in #434 — split out separately since the bugs predate this branch (pre-existing on `main`, just never exercised against a single-logit model before). This branch will be rebased on it once merged.

PR #434, opened 2026-08-11
fix: RMIA, BASE, and loss_trajectory attacks crash on single-logit (binary) models

fazelehh:
# Description

Fixes three pre-existing bugs in `AttackRMIA`, `AttackBASE`, and `AttackLossTrajectory` that crash when the target model has a single-logit (binary) output instead of a multi-class one. Split out of #423, where these were surfaced during review of the tabular/LoS webapp support — but the bugs themselves are independent of the webapp and already exist on `main`.

* **RMIA** — `_get_z_logits()`'s on-the-fly logit path used a bare `.squeeze()`, which collapses the num_classes axis along with the num_models axis when both are size 1. Fixed to `.squeeze(axis=0)`, matching the convention already used by `cache_logits()` elsewhere.
* **BASE** — `run_attack()`/`score_samples()` indexed raw logits assuming a column per class; a single-logit model has only column 0, so indexing label `1` raised `IndexError`. Fixed by routing through the existing `softmax_logits()` helper (same one RMIA already uses), which sigmoid-expands single-logit output before any indexing.
* **loss_trajectory** — `_prepare_mia_data()` fed Long-dtype labels into `BCEWithLogitsLoss`, which requires float targets. Fixed to cast based on the actual criterion in use (`isinstance(criterion, nn.BCEWithLogitsLoss)`), leaving `CrossEntropyLoss`'s Long requirement untouched.

None of these are regressions from binary support being added — they're old code paths (RMIA's on-the-fly branch was added in a recent, separate correctness fix; BASE and loss_trajectory's LOS-audit configs never even included those attacks before) that had simply never been exercised against a single-logit model until now.

## How Has This Been Tested?

* Full `leakpro/tests/mia_attacks/` suite: 47/47 pass, unchanged from `main` — confirms no regression to the existing multi-class path.
* Root-caused each crash against the actual tracebacks reported in #423's review (RMIA shape mismatch, BASE `IndexError`, loss_trajectory `RuntimeError`).

### DIFF
--- a/leakpro/attacks/mia_attacks/base.py
+++ b/leakpro/attacks/mia_attacks/base.py
@@ -6,14 +6,12 @@
 
 import numpy as np
 from pydantic import BaseModel, Field
-from scipy.special import (
-    log_softmax,
-    logsumexp,
-)
+from scipy.special import logsumexp
 from tqdm import tqdm
 
 from leakpro.attacks.mia_attacks.abstract_mia import AbstractMIA
 from leakpro.attacks.utils.shadow_model_handler import ShadowModelHandler
+from leakpro.attacks.utils.utils import softmax_logits
 from leakpro.input_handler.abstract_input_handler import AbstractInputHandler
 from leakpro.reporting.mia_result import MIAResult
 from leakpro.signals.signal import ModelLogits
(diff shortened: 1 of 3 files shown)


TheColdIce (approved):
Nice!

Issue #76, opened 2024-06-05
Label-only trajectory loss


johanos1:
# Issue
## Problem Description
The MIA data creation relies on logits and the knowledge distillation for the target teacher model relies only on the ground truth label.

## Expected Behavior
Utilize the hopskipjump to estimate the true class logit

## What Needs to be Done
Make use of hopskipjump instead

PR #440, opened 2026-08-13
fix: distill label-only loss trajectory on teacher labels, not ground truth

### LINKED ISSUE #146: Update loss trajectory label only with correct teacher
# Issue
## Problem Description
Look at todo in code

## Expected Behavior
We should not use ground truth but teacher model

## What Needs to be Done
Change the label only loss

run attack

fazelehh:
Closes #146.

## Problem

In label-only mode, the distillation student was trained with `CrossEntropyLoss` against the dataset's **ground-truth labels** (`distillation_model_handler.py`), so the distilled trajectory tracked the data rather than the teacher model being attacked. The `# TODO: I think this is wrong (teacher?)` on that line already suspected this.

## Fix

- Label-only distillation now targets the **teacher's predicted hard labels**: `argmax` of the teacher logits for multiclass, sign of the logit with `BCEWithLogitsLoss` for single-logit binary models (consistent with the binary handling added in #434).
- Ground-truth labels no longer enter distillation in either mode (the KL mode never used them).

## Tests

New `test_distillation_model_handler.py` distills a student from a constant-prediction teacher and asserts the student collapses onto the teacher's class instead of the dataset labels — for both the multiclass and single-logit paths. Both tests fail on the old code. `test_all_attacks_end_to_end.py -k loss_traj` passes.

Note: #76 (estimating the true class logit via HopSkipJump) is a further improvement on top of this and stays open.

https://claude.ai/code/session_01U4j7ifPMD8R2Mgd7om97F2

### DIFF
--- a/leakpro/attacks/utils/distillation_model_handler.py
+++ b/leakpro/attacks/utils/distillation_model_handler.py
@@ -10,7 +10,7 @@ import pickle
 import numpy as np
 import torch.nn.functional as F  # noqa: N812
 from torch import cat, cuda, device, save, sigmoid
-from torch.nn import CrossEntropyLoss, KLDivLoss, Module
+from torch.nn import BCEWithLogitsLoss, CrossEntropyLoss, KLDivLoss, Module
 from tqdm import tqdm
 
 from leakpro.attacks.utils.model_handler import ModelHandler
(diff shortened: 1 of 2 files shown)


henrikfo (approved):
Look great!

Issue #432, opened 2026-06-26
Too strict hashing?


henrikfo:
# Issue
## Problem Description
Not reusing shadow models when simply re-running leakpro

## Expected Behavior
Should reuse X shadow models in a simple re-run if X models have been trained.

## How Can It Be Tested or Reproduced
Run leakpro two times.

Issue #334, opened 2025-10-01
Shadow model datasplit


johanos1:
# Feature

## Desired Behavior / Functionality
When in offline mode, each data point should have the same number of in and out members

## What Needs to Be Done
Implement the data split

PR #441, opened 2026-08-13
fix: honor training_data_fraction when training shadow models

### LINKED ISSUE #345: Not reusing shadow models
# Issue
## Problem Description
When training shadow models and not using "training_data_fraction" = 0.5, e.g. 0.72, the shadow models is always retrained because they are trained with 0.5. 

## Expected Behavior
Should use "training_data_fraction" when selecting amount of data points to use when creating shadow models.

## What Needs to be Done
Make construct_balanced_assignments() compatible with other splits than 50/50

Run examples/mia/cifar_main.ipynb and change training_data_fraction to something else than 0.5

fazelehh:
Closes #345.

## Problem

`construct_balanced_assignments()` always assigned exactly half of the shadow population to every shadow model, so any `training_data_fraction` other than 0.5 was **silently ignored**. Worse: the cache signature records the fraction-derived `data_size`, while the models were actually trained on the half-population size — so with fraction ≠ 0.5 cached shadow models never matched their signature and every run retrained from scratch (the behavior reported in #345, and one of the two causes behind #432).

## Fix

- `construct_balanced_assignments()` takes `points_per_model` (the same `data_size` the cache signature records) and uses greedy least-loaded selection with random tie-breaking.
- Every model gets exactly the requested number of points, and each point's inclusion count stays within 1 across models — the balanced IN/OUT property (#334) now holds for **any** fraction, not just 0.5.
- Default stays at half the population, so existing caches for fraction 0.5 remain valid.

## Tests

- `test_construct_balanced_assignments_respects_points_per_model`: exact per-model sizes and ≤1 inclusion-count spread for fractions 0.3 / 0.5 / 0.72 / 1.0, plus rejection of degenerate sizes.
- `test_shadow_model_training_fraction_is_used_and_cached`: shadow models train on `int(0.75 * population)` points and a re-run with the same fraction reuses the cache instead of retraining. Both fail on the old code.
- Full `leakpro/tests/mia_attacks/` suite passes (49/49).

https://claude.ai/code/session_01U4j7ifPMD8R2Mgd7om97F2

### DIFF
--- a/leakpro/attacks/utils/shadow_model_handler.py
+++ b/leakpro/attacks/utils/shadow_model_handler.py
@@ -186,13 +186,18 @@ class ShadowModelHandler(ModelHandler):
 
         return all_indices, filtered_indices
 
-    def construct_balanced_assignments(self, m: int, n: int, seed: int = None) -> np.ndarray:
-        """Assigns each of m data points to approximately n/2 out of n datasets by partitioning.
+    def construct_balanced_assignments(self, m: int, n: int, points_per_model: int = None, seed: int = None) -> np.ndarray:
+        """Assigns each of m data points to n datasets of points_per_model points each, keeping per-point balance.
+
+        Every dataset gets exactly points_per_model points, and each point ends up
+        in an equal number of datasets (within 1), so each point has both IN and
+        OUT models regardless of the training fraction.
 
         Args:
         ----
             m (int): The number of data points.
             n (int): The number of datasets.
+            points_per_model (int, optional): Number of points per dataset. Defaults to m // 2.
             seed (int, optional): Random seed for reproducibility. Defaults to None.
 
         Returns:
@@ -203,19 +208,23 @@ class ShadowModelHandler(ModelHandler):
         if seed is not None:
             np.random.seed(seed)
 
+        if points_per_model is None:
+            points_per_model = m // 2
+        if not 0 < points_per_model <= m:
+            raise ValueError(f"points_per_model must be in [1, {m}], got {points_per_model}")
+
         A = np.zeros((n, m), dtype=np.uint8)  # noqa: N806
-        all_indices = np.arange(m)
+        inclusion_counts = np.zeros(m, dtype=np.int64)
 
-        for i in range(0, n - 1, 2):
-            permuted = np.random.permutation(all_indices)
-            half = m // 2
-            A[i, permuted[:half]] = 1       # First half to dataset i
-            A[i+1, permuted[half:]] = 1     # Second half to dataset i+1
-
-        if n % 2 == 1:
-            permuted = np.random.permutation(all_indices)
-            half = m // 2
-            A[n - 1, permuted[:half]] = 1   # Last dataset gets half for odd n
+        # Greedy least-loaded assignment: each dataset takes the points that are
+        # currently members of the fewest datasets (random tie-breaking), which
+        # keeps the per-point inclusion counts within 1 of each other.
+        for i in range(n):
+            tie_breaker = np.random.permutation(m)
+            order = np.lexsort((tie_breaker, inclusion_counts))
+            chosen = order[:points_per_model]
+            A[i, chosen] = 1
+            inclusion_counts[chosen] += 1
 
         return A
 
@@ -262,10 +271,11 @@ class ShadowModelHandler(ModelHandler):
             indices_to_use.append(next_index)
             next_index += 1
 
-        A = self.construct_balanced_assignments(len(shadow_population), num_models)  # noqa: N806
-        expected_size = len(shadow_population) // 2
-        if not np.all(np.sum(A, axis=1) == expected_size):
-            raise ValueError("Balanced shadow assignments must contain half of the shadow population per model")
+        # data_size is also what the cache signature records, so the trained size
+        # must match it exactly or cached models are never reused (issue #345).
+        A = self.construct_balanced_assignments(len(shadow_population), num_models, points_per_model=data_size)  # noqa: N806
+        if not np.all(np.sum(A, axis=1) == data_size):
+            raise ValueError("Balanced shadow assignments must contain data_size points per model")
         shadow_population = np.array(shadow_population)
         for i, indx in enumerate(indices_to_use):
             # Get dataloader
--- a/leakpro/tests/mia_attacks/utils/test_shadow_model_handler.py
+++ b/leakpro/tests/mia_attacks/utils/test_shadow_model_handler.py
@@ -327,3 +327,67 @@ def test_filter_rejects_shadow_models_from_different_population(image_handler: I
 
     assert all_indices == [0]
     assert filtered_indices == []
+
+
+def test_construct_balanced_assignments_respects_points_per_model(image_handler: ImageInputHandler) -> None:
+    """Every dataset must get exactly points_per_model points, with per-point counts balanced within 1."""
+    if ShadowModelHandler.is_created() is True:
+        ShadowModelHandler.delete_instance()
+    sm = ShadowModelHandler(image_handler)
+
+    m, n = 100, 7
+    for fraction in [0.3, 0.5, 0.72, 1.0]:
+        points_per_model = int(m * fraction)
+        A = sm.construct_balanced_assignments(m, n, points_per_model=points_per_model)
+        assert A.shape == (n, m)
+        assert np.all(A.sum(axis=1) == points_per_model)
+        inclusion_counts = A.sum(axis=0)
+        assert inclusion_counts.max() - inclusion_counts.min() <= 1
+
+    # Default keeps the historical behavior of half the population per model
+    A = sm.construct_balanced_assignments(101, 4)
+    assert np.all(A.sum(axis=1) == 101 // 2)
+
+    with raises(ValueError):
+        sm.construct_balanced_assignments(m, n, points_per_model=0)
+    with raises(ValueError):
+        sm.construct_balanced_assignments(m, n, points_per_model=m + 1)
+
+
+def test_shadow_model_training_fraction_is_used_and_cached(
+    image_handler: ImageInputHandler,
+    monkeypatch,
+    tmp_path,
+) -> None:
+    """Shadow models must train on training_fraction of the population and be reused for that fraction."""
+    shadow_config = ShadowModelConfig(**get_shadow_model_config())
+    image_handler.configs.shadow_model = shadow_config
+
+    if ShadowModelHandler.is_created() is True:
+        ShadowModelHandler.delete_instance()
+    sm = ShadowModelHandler(image_handler)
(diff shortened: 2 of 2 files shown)


leakpro/attacks/utils/shadow_model_handler.py
```diff
@@ -203,19 +208,23 @@ def construct_balanced_assignments(self, m: int, n: int, seed: int = None) -> np
...
+
+        # Greedy least-loaded assignment: each dataset takes the points that are
+        # currently members of the fewest datasets (random tie-breaking), which
+        # keeps the per-point inclusion counts within 1 of each other.
+        for i in range(n):
+            tie_breaker = np.random.permutation(m)
+            order = np.lexsort((tie_breaker, inclusion_counts))
+            chosen = order[:points_per_model]
+            A[i, chosen] = 1
+            inclusion_counts[chosen] += 1
```
henrikfo, line 227: The removal of forcing half-and-half splits are good and we should not enforce it. But this might create problems I think with LiRA offline for example, since data fraction can be > 0.5, lets say 0.7 and if we just train two models, then we might get some "ValueError("Score is NaN")". 
Should we have some filtering added to lira.py, somehting similar to whats in yoqo.py/dts.py. I think we had it before in lira.py but removed it seems. 

This will also "break" the BASE attack since it forces the shadow models to have 0.5, how should we handle that? (line 151 in base)

henrikfo: Similarly, and this is not anything directly added or changed in this PR, it is allowed to have "training_fraction=1.0" I believe. Should this be an issue or is it okay to have it like that? I'm thinking, do we have any attack that reasonably could use training_fraction=1.0? or should we force training_fraction<1.0?

leakpro/attacks/utils/shadow_model_handler.py
```diff
@@ -262,10 +271,11 @@ def create_shadow_models(
             indices_to_use.append(next_index)
             next_index += 1
 
-        A = self.construct_balanced_assignments(len(shadow_population), num_models)  # noqa: N806
-        expected_size = len(shadow_population) // 2
-        if not np.all(np.sum(A, axis=1) == expected_size):
-            raise ValueError("Balanced shadow assignments must contain half of the shadow population per model")
+        # data_size is also what the cache signature records, so the trained size
```
henrikfo, line 274: I really like this new code for balanced assignment. 

Should we incorporate cached models into this process? In the off chance that someone either quits, stop or their process crashes, I don understand what happens when the new training is started up, I don't think it will resume the same tie-breaker splits, but I think it would restart with a completely new one.

henrikfo (changes requested):
Some changes/ additions is needed to make sure all attacks for as intended I think.

Also added some thoughts on making these leakpro changes a little bit more robust, not needed for approval though! :)

leakpro/attacks/utils/shadow_model_handler.py
fazelehh: You were right on both counts, and the LiRA one is worse than a possible NaN: it is guaranteed in a predictable regime.

**Why it happens.** Greedy least-loaded keeps per-point membership counts within 1 of `num_shadow_models * training_data_fraction`. So a point has no OUT model when the fraction exceeds `1 - 1/num_shadow_models`, and no IN model when it falls below `1/num_shadow_models`. Your example is exactly the first case. Simulated at m=100:

| num_shadow_models | fraction | OUT counts | offline scorable | online scorable |
|---|---|---|---|---|
| 2 | 0.7 | 0-1 | 60/100 | 60/100 |
| 2 | 0.3 | 1-2 | 100/100 | 60/100 |
| 2 | 0.5 | 1-1 | 100/100 | 100/100 |
| 4 | 0.5 | 2-2 | 100/100 | 100/100 |

So at 2 models and 0.7, 40% of audit points hit `ValueError("Score is NaN")`.

**Fix (a3f5059).** Added the filtering you asked for, as a shared predicate `audit_points_with_reference_models` in `shadow_model_handler.py` rather than a third copy of the `yoqo.py`/`dts.py` block. Offline requires an OUT model, online requires both sides. LiRA and MS-LiRA both use it; MS-LiRA had this on its online path only, and its offline path needed it too.

One deliberate difference from yoqo/dts: dropping points is loud, not silent. Under balanced sampling a non-empty drop set means the fraction is wrong for the model count, so the drop is structural rather than incidental and the warning names the counts and the fraction that produced them. Silently auditing 60% of the points would understate the audit.

**BASE (c82101b).** The assert is not just incompatible with this PR, it is already broken on main. It demands exactly `num_shadow_models // 2` OUT models per point, but the real requirement is only that the count is positive, since the threshold is `logsumexp(out_logits) - log(n_out_models)`, a log mean over whatever OUT models exist. Ran main's assignment code against main's assert:

| num_shadow_models | OUT counts | assert passes |
|---|---|---|
| 2 | 1-1 | yes |
| 3 | 1-2 | **no** |
| 4 | 2-2 | yes |
| 5 | 2-3 | **no** |

Offline BASE fails today for any odd `num_shadow_models`. Replaced both copies of the assert with a check that no scored point lacks an OUT model, raising an error naming the fraction and the model count. I filtered in LiRA but raise in BASE on purpose: `score_samples` is driven by RaMIA over range clusters, and dropping points there would silently change the clusters.

The odd-`num_shadow_models` half of that is a pre-existing bug and I would normally split it out. I kept it here because it is the same assert that blocks this PR's feature, and fixing half of it makes no sense. Happy to pull it into its own PR against main if you would rather keep this branch narrow.

fazelehh: Agreed, and it should be forbidden. No shadow-model attack in the repo can use 1.0: it trains every shadow model on every point, so no point has an OUT reference model and LiRA, RMIA, BASE and the rest have nothing to calibrate against. It is not a weak configuration, it is a broken one. 0.0 is equally meaningless, and before this PR it silently trained on half the population anyway.

Fixed rather than filed (d62b56e), in two places so it fails early and cannot be bypassed:

- Config validation is now `gt=0.0, lt=1.0` in `base.py`, `dts.py`, `lira.py`, `multi_signal_lira.py`, `oslo.py`, `ramia.py`, `rmia.py`, `yoqo.py`.
- `create_shadow_models` rejects the resulting `data_size` of 0 or full population, so any future caller is covered even if it does not go through one of those configs.

`qmia.py` deliberately keeps `ge/le`: its `training_data_fraction` subsamples the attack data for the quantile regressor and never drives shadow-model membership, so 1.0 means "use all of it" there and is legitimate.

Worth knowing that `lt=1.0` alone is not sufficient. As in the thread above, the fraction also has to stay inside `[1/num_shadow_models, 1 - 1/num_shadow_models]` for every point to keep both sides, and that bound depends on the model count, so it cannot live in a per-field validator. That is what the LiRA warning and the BASE error now report at runtime.

leakpro/attacks/utils/shadow_model_handler.py
fazelehh: Your instinct was right, and the code was worse than "starts with a new tie-breaker". Cached models were not consulted at all.

The old flow built `A` for all `num_models` rows, then iterated `for i, indx in enumerate(indices_to_use)` where `indices_to_use` holds only the models still to be trained. So with 8 cached models and `num_shadow_models=16`, a freshly drawn design's first 8 rows went to the 8 new models, and the 8 cached models' actual memberships never entered the calculation. The result is two independent balanced designs whose union is only balanced within 2, not the within-1 property #334 promises. Same thing on a crash-resume, same thing on any increase of `num_shadow_models`.

Fix (89138f4): `construct_balanced_assignments` takes `prior_inclusion_counts`, and `create_shadow_models` seeds it from the cached models' stored `train_indices` (read back via `get_in_indices_mask`) and generates rows only for the models it is about to train.

Replaying the tie-breaker would have been the other option, and it is the weaker one. It needs the RNG state persisted and the models trained in the same order, and it breaks the moment `num_shadow_models` changes. Seeding from what the cached models actually contain needs no stored state and handles the "asked for more models" case identically to the crash-resume case.

Two tests cover it: one that a seeded pass is forced onto exactly the points a cached model missed, and one that trains 2 models, then asks for 4, and asserts the union of all 4 still has per-point counts within 1. The second fails on the old code.

fazelehh:
@henrikfo All four points addressed. Every one still reproduced against current `main`, so none had aged out. Details are in the three threads; summary here.

Branch first merged `origin/main` (it was cut before the `get_device`/seed work landed), then four commits:

| Commit | Your point | What changed |
|---|---|---|
| 89138f4 | cached models and resume | `construct_balanced_assignments` takes `prior_inclusion_counts`; `create_shadow_models` seeds it from cached models' `train_indices` and generates rows only for new models |
| a3f5059 | LiRA NaN with fraction > 0.5 | Shared `audit_points_with_reference_models` predicate; LiRA and MS-LiRA drop unscorable points and warn with the counts and the fraction |
| c82101b | BASE forces 0.5 | Both `num_shadow_models // 2` asserts replaced with a positive-OUT-count check |
| d62b56e | is 1.0 allowed | `gt=0.0, lt=1.0` across the eight shadow-model attack configs, plus the same guard inside the handler |

Two things worth your attention before approving.

**I fixed a pre-existing bug inside this PR.** The BASE assert already fails on `main` for any odd `num_shadow_models`, independent of this branch. I kept the fix here because it is the same assert that blocks the feature, but say the word and I will split it into its own PR against `main`.

**LiRA drops points, BASE raises.** Deliberate asymmetry. `score_samples` is driven by RaMIA over range clusters, and silently dropping points there would change the clusters rather than just the sample size. Pushed back slightly on the yoqo/dts precedent in one respect: the drop is logged as a warning naming the configuration, because under balanced sampling a non-empty drop set means the fraction is wrong for the model count, so it is structural rather than incidental. A silently smaller audit set understates the result.

Not done, flagging rather than fixing: `yoqo.py` and `dts.py` still carry two verbatim copies of the filtering block. Correcting myself here, since I first wrote that switching them over was mechanical and had not checked. It is not, in general. Neither attack computes a per-point OUT statistic: yoqo uses the mask as per-model loss weights and sets them to all-ones offline, and dts uses it only for the online filter and an offline log line. So neither can hit the empty-slice NaN the predicate exists to prevent, and their online-only filtering is correct for them. A behaviour-preserving dedup is still available, about six lines, but only with the call kept inside the `if self.online` branch. Adopting the predicate the way the sentence above implied would make them drop points they can legitimately score. Worth a separate PR at most.

Verification: 440 passed, ruff clean, on `d62b56e` in `leakpro_py311`. Eleven new tests, including two that fail on the old code. Two failures excluded as environment issues, both confirmed identical on `origin/main`: four collection errors from missing `transformers`/`diffusers` extras, and `test_aux_file_path`, which hardcodes the repo directory name and fails in any worktree.

henrikfo (approved):

PR #447, opened 2026-08-18
PETRecipe: one shared DP-SGD training path

fazelehh:
Stacked on #438 (base is `feature/pet-optimization`; GitHub retargets this to main automatically once #438 merges).

## What

Adds `leakpro/optimization/training.py`: the structured training contract (`PETRecipe`) and the single shared Opacus path (`train_with_dpsgd`, `build_campaign_fns`) that replaces the DP-SGD loop previously duplicated in every campaign example. All three examples (LOS-LR, CIFAR, LOS GRU-D) are ported onto it and shrink to declaring a recipe; the standalone GRU-D script is superseded by the recipe-ported one.

A `PETRecipe` is the plan's structured contract: instead of a monolithic `train(data) -> model`, the user supplies factories the optimizer can intervene in (`make_model`, `make_optimizer`, `make_loader`, criterion, epochs). Each factory receives the sampled knob config and reads what it needs, which is what makes the next PET cheap — a regularization bundle is just `make_model` reading `dropout` and `make_optimizer` reading `weight_decay`, with no interface change. Exotic loops that do not fit keep the escape hatch of writing the three campaign callables directly.

## Hardening beyond the refactor

- **`make_opacus_compatible()`** — ModuleValidator pass (BatchNorm → GroupNorm), disables in-place activations, and rewrites torchvision residual blocks whose in-place `out += identity` ModuleValidator cannot see because it lives in `forward` rather than a submodule. Without this, any BatchNorm architecture — including both webapp image presets — fails at `make_private()`.
- **The privacy accountant is a parameter**, defaulting to `"prv"` (was hard-coded `"rdp"`). RDP is the looser bound, so identical noise reported a larger ε here than via the webapp's DP-SGD path. This changes reported ε for existing campaigns (downward, for identical noise); empirical attack numbers are unaffected.
- **The accuracy metric thresholds single-logit binary heads at 0** instead of `argmax` — `argmax` over one column is always 0, silently yielding 0% or 100%.

## Review fixes applied

An adversarial review of this PR produced 10 findings; all are addressed in `18db5a2a`. The theme was silent behaviour changes versus the per-example code this PR replaced.

- **`train_with_dpsgd` fails closed.** A missing or misspelled `noise_multiplier` silently trained a fully non-private model, from a function whose only purpose is DP-SGD. The key is now required; non-private runs pass an explicit `0.0`.
- **`make_opacus_compatible` runs on both branches.** Rewriting BatchNorm → GroupNorm only for private configs left the noise=0 anchor a structurally different model, so its utility gap mixed the cost of DP noise with the cost of an architecture change — the frontier's utility axis was not comparable end to end. Documented consequence: the submitted architecture is not necessarily what trains.
- **Full mimicry is enforced, not silently downgraded.** `n_ref_train` used `min(target, ref_pool)`, so a short reference pool quietly trained weaker references and biased the audit optimistic. A short pool is now an error — which is what the old per-example code did loudly.
- **`_patch_residual_blocks` only rewrites blocks still using the stock `forward`.** Subclasses with their own forward (SE, attention) were being silently replaced by the vanilla residual path, changing the model rather than making it Opacus-safe. Such blocks are now left alone with a warning.
- **The `auc` utility metric refuses multiclass output** instead of `reshape(-1)`-ing it into a meaningless number.
- **The accountant travels with the epsilon** in `campaign_extras`. The `prv` default is deliberate, but PRV and RDP epsilons are not comparable — measured 4.22 vs 4.87 for identical noise — so a record that does not name its accountant cannot safely be compared with another run's.

Examples: GRU-D regains the `max_physical_batch=128` cap the pre-PR code set deliberately (it is the memory-heaviest target in the repo; the shared 256 default would OOM mid-campaign) and gains the small-dataset split guard its LR sibling already had. The LOS campaign imports `target_models.LR` instead of redeclaring an identical `LRSigmoid`, so it trains the same class the audited target used.

One earlier test asserted the now-incorrect behaviour (BatchNorm surviving the non-private path) and has been inverted to assert architecture consistency instead.

## Verification

- `pytest leakpro/tests/test_optimization/` — 47 tests pass.
- `ruff check .` clean.
- The DP-SGD compatibility tests fail with the fix disabled (checked), so they are real regression tests rather than vacuous ones.

The webapp's near-verbatim copy of the Opacus-compat block is removed in #448, which now calls this shared helper.

https://claude.ai/code/session_01QBPG7MprEqhqG95fFURk8h

### DIFF
--- a/examples/mia/LOS/pet_optimization/run_campaign.py
+++ b/examples/mia/LOS/pet_optimization/run_campaign.py
@@ -72,22 +72,18 @@ def load_splits(seed: int = 0) -> dict:
     }
 
 
-class LRSigmoid(nn.Module):
-    """Logistic regression emitting probabilities (matches the LOS example's LR + BCELoss)."""
-
-    def __init__(self, input_dim: int) -> None:
-        super().__init__()
-        self.linear = nn.Linear(input_dim, 1)
-
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
-        return torch.sigmoid(self.linear(x))
-
-
 def make_recipe(splits: dict, epochs: int) -> PETRecipe:
-    """The LOS training recipe: logistic regression + Adam + BCE."""
+    """The LOS training recipe: logistic regression + Adam + BCE.
+
+    The model is the example's own ``target_models.LR`` — the same class the
+    audited target was trained with, so the campaign measures that model rather
+    than a look-alike.
+    """
+    from target_models import LR  # noqa: PLC0415 — EXAMPLE_DIR is on sys.path above
+
     input_dim = splits["x"].shape[1]
     return PETRecipe(
-        make_model=lambda config: LRSigmoid(input_dim),
+        make_model=lambda config: LR(input_dim),
         make_optimizer=lambda params, config: optim.Adam(params, lr=config["learning_rate"]),
         make_loader=lambda indices, config: DataLoader(
             TensorDataset(splits["x"][indices], splits["y"][indices]),
--- a/examples/mia/LOS/pet_optimization/run_campaign.py
+++ b/examples/mia/LOS/pet_optimization/run_campaign.py
@@ -5,15 +5,9 @@
 """PET optimization campaign on the MIMIC LOS logistic-regression target.
 
 Traces the utility-vs-attack-success frontier for DP-SGD on the LOS binary
-classification task, using the leakpro.optimization core module.
-
-- Knobs (joint search, per the frontier plan): noise multiplier, clipping
-  norm, learning rate, batch size. Epochs are fixed.
-- Utility: AUC on a held-out test split.
-- Attack: matched-reference MIA (RMIA-style calibration with 1-2 reference
-  models trained with the SAME configuration as the candidate — full mimicry).
-  Signal: logit-scaled confidence of the target minus the reference mean.
-- Loop metric: TPR @ FPR = 1% (powered proxy; tail FPRs are validation-only).
+classification task. The training loop, attack and utility evaluation all come
+from the shared ``PETRecipe`` path in ``leakpro.optimization`` — this file only
+declares the recipe (LR model, Adam, BCE loss) and the data splits.
 
 Usage:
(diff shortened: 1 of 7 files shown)


fazelehh:
Addressed the code-review findings on this PR. The theme was silent behaviour changes versus the per-example code this PR replaced.

**Correctness**
- `train_with_dpsgd` **fails closed**: a missing or misspelled `noise_multiplier` silently trained a fully non-private model. The key is now required; non-private runs pass an explicit `0.0`.
- `make_opacus_compatible` runs on **both** branches. Rewriting BatchNorm→GroupNorm only for private configs left the noise=0 anchor a structurally different model, so its utility gap mixed the cost of DP noise with the cost of an architecture change — the frontier's utility axis was not comparable end to end.
- **Full mimicry is enforced, not silently downgraded**: `n_ref_train` used `min(target, ref_pool)`, so a short reference pool quietly trained weaker references and biased the audit optimistic. Now an error, as the old per-example code was.
- `_patch_residual_blocks` only rewrites blocks still using the stock `forward`; subclasses with a custom forward (SE/attention) were being silently replaced by the vanilla residual path.
- The `auc` utility metric refuses multiclass output instead of `reshape(-1)`-ing it into a meaningless number.
- The **accountant travels with the epsilon** in `campaign_extras`. The `prv` default is deliberate (it matches the webapp's DP-SGD path), but PRV and RDP epsilons are not comparable — measured 4.22 vs 4.87 for identical noise — so a record that does not name its accountant cannot safely be compared with another run's.

**Examples**
- GRU-D regains `max_physical_batch=128` (it is the memory-heaviest target here; the shared 256 default would OOM mid-campaign) and gains the small-dataset split guard its LR sibling already had.
- The LOS campaign imports `target_models.LR` instead of redeclaring an identical `LRSigmoid`, so it trains the same class the audited target used.

**Also**: one of my earlier tests asserted the now-incorrect behaviour (BatchNorm surviving the non-private path) and has been inverted to assert architecture consistency. The webapp's duplicate Opacus-compat block is removed in #448, which now calls this shared helper.

47 tests pass on this branch, ruff clean.

PR #425, opened 2026-05-26
Add HPU support to leakpro

Muhaddisabarat:
# Description

Summary of changes

* Many files have been modified mostly based on import get.device 
* 2 files added.
  a) leakpro/utils/hardware_setup.py (Hardware selection order )
     1. Habana Gaudi HPU (if ``habana_frameworks.torch`` is installed and reports an available device).
     2. NVIDIA CUDA (if ``torch.cuda.is_available()``).
     3. CPU. )
     
   b) leakpro/utils/device.py (Hardware detection utility for selecting the runtime device).

---

**Related issues:** #442 (HPU support for the time-series MIA example) — partially addressed by this PR (see PR comments for status); left open, not closed, since hardware validation and dedicated docs for the TS2Vec CPU fallback are still outstanding.

PR #451, opened 2026-08-19
fix: apply audit random_seed at audit start and before every attack

### LINKED ISSUE #325: Random seeding overhaul
Right now setting the random_seed parameter in the audit config does nothing, the only part of the code base that seeds everything is using the optuna seed at the start of optuna trials.
Ideally it should seed everything when:
- Starting an audit
- Starting an attack (preferred for reproducability when some results are already run)

Also, the current seed_everything function used by optuna is problematic in two ways:
- Deterministic cudnn is not necessary for most models and usually slows down training/inference by some 30% (in some cases 100x slower)
- Setting cudnn in benchmarking mode has nothing to do with seeding, but is generally good for performance

fazelehh:
Closes #325.

## Validation of the issue

Confirmed on current main: `random_seed` is defined in `AuditConfig` (`schemas.py`) but consumed nowhere — the only `seed_everything` caller is Optuna, with its own separate seed. Also confirmed the issue's second complaint: `seed_everything` forces `torch.backends.cudnn.deterministic = True` and `benchmark = False`, which is a performance trap unrelated to seeding. Both parts valid.

## Fix

- `AttackScheduler.__init__` seeds with `audit.random_seed` before constructing attacks (audit start), falling back to the `AuditConfig` schema default when the config lacks the field.
- `run_attacks` re-seeds with the same value **before every attack**, so each attack sees identical random state regardless of which earlier attacks ran live or were loaded from cache — the reproducibility-with-cached-results case the issue calls out.
- A `random_seed` that is present but not an int (a config typo such as `"42"`, or `true`, which is a `bool` and therefore an `int` subclass) logs a warning naming the rejected value and the fallback, instead of silently defaulting. A genuinely absent field stays silent, since falling back is correct there.
- `seed_everything` no longer touches the cuDNN flags; it only seeds torch / numpy / random (+ CUDA and HPU).

### Scope note on the cuDNN change

Dropping the cuDNN flags is the second of the two problems named in #325, not a change bundled in for convenience. It is nevertheless a behaviour change reaching beyond the `audit.random_seed` path, so, explicitly: **every** caller of `seed_everything` loses deterministic cuDNN and keeps benchmark mode. That is Optuna's per-trial seeding (`hyperparameter_tuning/optuna.py`), the GIA examples (`examples/gia/*/main.py`), and now the audit path added here. Runs on GPU are no longer bit-exact across repeats; anyone who needs that sets `torch.backends.cudnn.deterministic = True` themselves. This is the intended trade — the issue reports the flag costing up to 100x on some models.

## Merge with main

Main has since merged #425, which rewrote `seed_everything` to add HPU (Habana) seeding and added `leakpro/tests/utils/test_seed.py`. Main is merged in here; the conflict resolution keeps main's CUDA/HPU seeding intact and removes only the two cuDNN lines. This branch's `leakpro/tests/test_seed.py` would have been a second test module with the same basename covering the same unit, so its tests were folded into `leakpro/tests/utils/test_seed.py` and the duplicate removed.

## Tests

In `leakpro/tests/utils/test_seed.py`: seed reproducibility across numpy/torch/random, cuDNN flags untouched, scheduler seeds once at construction and once per attack with the configured seed (spy on `seed_everything`), the schema-default fallback, and the warn-on-non-int path. `pytest leakpro/tests`: 433 passed. (One unrelated failure, `test_aux_file_path`, hardcodes `/LeakPro/` in an expected path and fails in any worktree not named `LeakPro`; four modules fail collection on optional deps absent from the local env.)

https://claude.ai/code/session_01JWoQzXWoysSRUxQUMh8q6q

### DIFF
--- a/leakpro/attacks/attack_scheduler.py
+++ b/leakpro/attacks/attack_scheduler.py
@@ -7,8 +7,10 @@
 from pathlib import Path
 
 from leakpro.input_handler.abstract_input_handler import AbstractInputHandler
+from leakpro.schemas import AuditConfig
 from leakpro.utils.import_helper import Any, Dict, Self
 from leakpro.utils.logger import logger
+from leakpro.utils.seed import seed_everything
 
 
 class AttackScheduler:
@@ -30,6 +32,12 @@ class AttackScheduler:
 
         """
         configs = handler.configs
+
+        # Seed once at audit start so attack construction (e.g. audit-data sampling)
+        # is reproducible. Non-int guard keeps DotMap-based test configs working.
+        seed = getattr(configs.audit, "random_seed", None)
+        self.random_seed = seed if isinstance(seed, int) else AuditConfig.model_fields["random_seed"].default
+        seed_everything(self.random_seed)
 
         # Create factory
         attack_type = configs.audit.attack_type
@@ -103,6 +111,9 @@ class AttackScheduler:
         """Run the attacks and return the results."""
         results = []
         for attack_obj, attack_type in zip(self.attacks, self.attack_names):
+            # Re-seed before every attack so each attack sees the same random state
+            # regardless of which earlier attacks ran or were loaded from cache.
+            seed_everything(self.random_seed)
             run_with_optuna = use_optuna and attack_obj.optuna_params > 0
 
             # If Optuna is used, the attack should not be loaded even if it already exists
--- a/leakpro/utils/seed.py
+++ b/leakpro/utils/seed.py
@@ -10,11 +10,16 @@ import torch
 
 
 def seed_everything(seed: int) -> None:
-    """Set the seed for different libraries."""
+    """Set the seed for different libraries.
+
+    Deliberately does not touch the cuDNN flags: forcing
+    torch.backends.cudnn.deterministic slows training/inference
+    substantially, and benchmark mode is a performance setting unrelated
+    to seeding (issue #325). Users who need bit-exact GPU reproducibility
+    can set those flags themselves.
+    """
     torch.manual_seed(seed)
     torch.cuda.manual_seed(seed)
     torch.cuda.manual_seed_all(seed)
     np.random.seed(seed)
     random.seed(seed)
-    torch.backends.cudnn.deterministic = True
-    torch.backends.cudnn.benchmark = False
--- a/leakpro/tests/test_seed.py
+++ b/leakpro/tests/test_seed.py
@@ -1,0 +1,94 @@
+#
+# Copyright 2023-2026 Lindholmen Science Park AB
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Test seeding: seed_everything behavior and audit/attack seeding in the scheduler."""
+
+import random
+
+import numpy as np
+import torch
+from dotmap import DotMap
+from pytest import MonkeyPatch
+
+from leakpro.attacks.attack_scheduler import AttackScheduler
+from leakpro.tests.input_handler.image_input_handler import ImageInputHandler
+from leakpro.utils.seed import seed_everything
+
+
+def test_seed_everything_is_reproducible() -> None:
+    """The same seed must reproduce the same numpy, torch and random draws."""
+    seed_everything(1234)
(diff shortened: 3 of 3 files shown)


leakpro/utils/seed.py
```diff
@@ -10,11 +10,16 @@
 
 
 def seed_everything(seed: int) -> None:
-    """Set the seed for different libraries."""
+    """Set the seed for different libraries.
```
Muhaddisabarat, line 13: Removing the cuDNN determinism/benchmark flags here affects every other caller of `seed_everything()` in the repo — the GIA examples (`examples/gia/*/main.py`) and Optuna's per-trial seeding in `hyperparameter_tuning/optuna.py` — not just the audit `random_seed` path this PR is fixing. The rationale referencing #325 makes sense, but since it's a behavior change reaching outside this PR's stated scope, can you confirm it's intentionally in scope for #325 (not just bundled in for convenience)? Worth a line in the PR description either way.

Muhaddisabarat (commented):
Reviewed the audit-seeding fix — the core logic (seed once at audit start, re-seed before every attack, fallback to the schema default when the config value is missing/non-int) is correct and well covered by the new tests in `test_seed.py`. I ran `test_seed_everything_is_reproducible` and `test_seed_everything_leaves_cudnn_flags_alone` locally and both pass. Left two non-blocking comments inline below. Nothing here looks blocking — happy to see this merged once the cuDNN-scope question is answered.

leakpro/attacks/attack_scheduler.py
```diff
@@ -31,6 +33,12 @@ def __init__(
         """
         configs = handler.configs
 
+        # Seed once at audit start so attack construction (e.g. audit-data sampling)
+        # is reproducible. Non-int guard keeps DotMap-based test configs working.
+        seed = getattr(configs.audit, "random_seed", None)
+        self.random_seed = seed if isinstance(seed, int) else AuditConfig.model_fields["random_seed"].default
```
Muhaddisabarat, line 39: When `configs.audit.random_seed` is present but not an int (e.g. a config typo — string or float), this silently falls back to the schema default (42) with no warning. Might be worth a `logger.warning` here so a bad seed value doesn't go unnoticed.

leakpro/utils/seed.py
fazelehh: Intentionally in scope — it's the issue's own second ask, not something bundled in for convenience. #325 lists two problems, and the cuDNN flags are one of them, verbatim:

> - Deterministic cudnn is not necessary for most models and usually slows down training/inference by some 30% (in some cases 100x slower)
> - Setting cudnn in benchmarking mode has nothing to do with seeding, but is generally good for performance

You're right that the blast radius is wider than the description said, though: it named only Optuna. I've updated the PR description to spell out that the GIA examples (`examples/gia/*/main.py`) and Optuna's per-trial seeding both lose deterministic cuDNN as a side effect, and that anyone needing bit-exact GPU kernels now sets the flags themselves.

One thing to re-check on your side: main has since merged #425, which rewrote `seed_everything` to add HPU seeding. I've merged main in and resolved the conflict so that main's HPU logic is kept intact and only the two cuDNN lines are dropped.

leakpro/attacks/attack_scheduler.py
fazelehh: Agreed, fixed in db2c148f. A bad value now logs the value it rejected and the seed it fell back to, so an audit can't quietly report a seed it never applied.

Two details that came out of implementing it:

- **Absent stays silent.** A DotMap returns an *empty DotMap* for a missing key rather than raising, so "missing" and "present but wrong" both fail `isinstance(seed, int)`. The warning is gated on the value not being `None` or an empty mapping — falling back is the correct, expected behaviour when the field genuinely isn't there, and warning on it would fire on every DotMap-based test config.
- **Booleans excluded from the int branch.** `bool` is a subclass of `int`, so `random_seed: true` would have seeded with 1 and passed the guard silently. It now takes the warn-and-default path.

Covered by `test_warns_and_defaults_on_non_int_seed`.

Issue #453, opened 2026-08-26
Critical: unauthenticated RCE in webapp backend (pickle deserialization on upload)


mkultraware:
# Unauthenticated RCE in the webapp backend via pickle deserialization on upload

## Summary

The webapp backend (`leakpro/webapp/backend`, commit `f52bb65`) accepts dataset uploads through `POST /jobs/{id}/upload/data` with no authentication, then deserializes them with `pickle.load()`. Pickle deserialization of untrusted input is arbitrary code execution. A crafted `.pkl` file runs attacker-controlled shell commands as the server process user.

Severity: critical, CVSS 3.1 9.8 (AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H)
Auth required: none
CORS: `allow_origins=["*"]`
Status: reproduced locally against an unmodified clone. Not tested against any deployed instance.

## Vulnerable path

`upload_data()` calls `inspector.inspect()`, which calls `inspector._try_load()`, which runs `pickle.load(f)` on the uploaded file. Same sink appears in `inspector.py` and `main.py`.

A second unauthenticated code execution path exists in `checker.run_check()`, which executes a user-uploaded `arch.py` through `subprocess` and `spec.loader.exec_module`.

## Impact

Full remote code execution as the backend service account, on any deployment where the port is reachable. This is a data privacy assessment tool built to handle models trained on sensitive data, so a compromised backend on a live instance would be a serious problem. I have not verified whether a live instance exists or what data it handles.

## Disclosure timeline

August 19, 2026: private disclosure sent to the maintainers with a seven day acknowledgment window.
August 26, 2026: no response received.
CVE reserved through MITRE CNA-LR: CAN-2026-2036101.

Posting this publicly now, after private disclosure went unanswered.

## Fixes

Do not run `pickle.load()` on untrusted uploads. Use JSON or Parquet, or a signed and allowlisted schema.
Add authentication to the backend.
Restrict CORS to known origins.
Do not execute user-supplied Python.
Confine every path-accepting endpoint to the job directory.

## Proof of concept

A working PoC exists and was run locally against an unmodified clone, confirming code execution as the server process. It is not included here while the bug is unpatched. Happy to send it to the maintainers directly for verification.

fazelehh:
This is now fully fixed on `main`. Thanks @mkultraware for the report and the constructive follow-up.

[...]

PR #455, opened 2026-09-02
fix(webapp): require authentication and validate paths on the backend API

fazelehh:
**Security fix** for the unauthenticated RCE reported in #453.

The webapp backend deserializes user model files and executes user-supplied Python by design — that's how it audits a model. This PR makes those operations reachable only by an authenticated local operator:

- **Bearer-token auth** on all `/jobs` routes (auto-generated and logged when `LEAKPRO_WEBAPP_TOKEN` is unset, so local dev stays a copy-paste)
- **CORS origin allowlist** replacing `allow_origins=["*"]`
- **WebSocket origin + token check** before `accept()` (WebSockets bypass CORS and cannot carry auth headers)
- **Loopback-only Docker binding** — the shipped compose previously published the backend on `0.0.0.0:8000`
- **Path-component validation**: `model_name` / `TrainParams.name` were joined into filesystem paths unvalidated; since `Path("/a/b") / "/etc"` is `/etc`, this was an arbitrary-file-write primitive, not just traversal
- **Confinement of the six server-side `*-path` endpoints** to configured data roots (`LEAKPRO_WEBAPP_DATA_ROOTS`; bundled examples keep working)
- **A genuinely restricted unpickler** for metadata validation; the previous `_SafeUnpickler` resolved `os.system` as readily as a real class and is renamed `_LenientUnpickler` with its limits documented
- Loaders try `torch.load(weights_only=True)` / `np.load(allow_pickle=False)` before falling back; upload size caps; bounds on `TrainParams`; tracebacks logged instead of returned; `/start` duplicate-submit race closed; `.dockerignore` so job data and `.git` stop being baked into images

**Verification:** unauthenticated → 401, wrong token → 401, cross-site origin → 403, `model_name` traversal/absolute path → 400 with no file written, `/data-path /etc/passwd` → 400, legitimate flow → 200. A canary pickle that provably executes under plain `pickle.load` is inert under the restricted unpickler. 11 new regression tests (`leakpro/tests/test_webapp_security.py`); existing suite passes (254); frontend `tsc` clean. Threat model documented in `leakpro/webapp/SECURITY.md`.

**Follow-ups, deliberately kept out of this diff to keep the security change auditable:** hardening cleanups (PR 2) and safe dataset formats — Parquet/`.npz` — implementing the format recommendation from #453 (PR 3).

Reported by **@mkultraware** (using Accuretta). Thanks for the responsible disclosure.

Refs #453 — the issue stays open until the safe-format PR lands; authentication is the accepted mitigation for the pickle sink in this PR.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,39 @@
+# The webapp image is built from the repo root (see leakpro/webapp/Dockerfile.backend).
+# Without this file, `COPY . .` bakes real audit data and git history into every
+# image layer — including anything pushed to a registry.
+
+# Version control
+.git
+.gitignore
+
+# Real job data, uploaded datasets and trained models
+webapp_jobs/
+webapp_*_files/
+leakpro_output/
+leakpro_outpu*/
+
+# Local environments and caches
+.venv/
+venv/
+**/__pycache__/
+*.py[cod]
+.pytest_cache/
+.ruff_cache/
+.mypy_cache/
+*.egg-info/
+
+# Node
+**/node_modules/
+**/dist/
+**/.vite/
+
+# Notebooks and scratch files
+.ipynb_checkpoints/
+*.swp
+*.swo
+*~
+
+# Local configuration
+.env
+.env.*
+CLAUDE.local.md
--- a/examples/mia/celebA_HQ/webapp/README.md
+++ b/examples/mia/celebA_HQ/webapp/README.md
@@ -16,10 +16,15 @@ cd leakpro/webapp/frontend && npm install
 
 **Backend** (from repo root):
 ```bash
-uvicorn leakpro.webapp.backend.main:app --reload --port 8000
+uvicorn leakpro.webapp.backend.main:app --reload --host 127.0.0.1 --port 8000
 # To use a different port:
-uvicorn leakpro.webapp.backend.main:app --reload --port 8001
+uvicorn leakpro.webapp.backend.main:app --reload --host 127.0.0.1 --port 8001
 ```
+
+The backend prints an API token at startup; paste it into the UI when prompted.
+Set `LEAKPRO_WEBAPP_TOKEN` to pin your own instead. Keep the bind address on
+`127.0.0.1`: this backend loads user model files and runs user-supplied Python
+by design. See [`leakpro/webapp/SECURITY.md`](../../../../leakpro/webapp/SECURITY.md).
 
 **Frontend** (from `leakpro/webapp/frontend/`):
 ```bash
--- a/leakpro/tests/test_webapp_security.py
+++ b/leakpro/tests/test_webapp_security.py
@@ -10,6 +10,7 @@ unauthenticated access, cross-site requests, path traversal through
 metadata deserialization.
 """
 
+import contextlib
 import os
 import pickle
 import tempfile
@@ -17,8 +18,8 @@ from pathlib import Path
 
 import pytest
 
-# The webapp is optional tooling with its own dependencies (see the webapp
-# READMEs); skip these tests entirely where fastapi is not installed, e.g. CI.
+# Webapp tests: run wherever the webapp extras are installed
+# (pip install -e ".[webapp]"); skipped elsewhere.
 pytest.importorskip("fastapi", reason="webapp extras (fastapi) not installed")
 
 from fastapi import HTTPException  # noqa: E402
@@ -67,6 +68,20 @@ class TestPathValidation:
         assert safe_suffix("../../evil") == ""
         assert safe_suffix(None) == ""
 
+    def test_trailing_newline_rejected(self: Self) -> None:
+        """`$` under match() accepts a trailing newline; fullmatch must not."""
+        from leakpro.webapp.backend.security import safe_name, safe_suffix
+
+        with pytest.raises(HTTPException):
+            safe_name("model1\n")
+        assert safe_suffix("x.pkl\n") == ""
+
+    def test_non_ascii_token_is_invalid_not_error(self: Self) -> None:
+        """compare_digest raises TypeError on non-ASCII str; must return False."""
+        from leakpro.webapp.backend.security import token_is_valid
+
+        assert token_is_valid("\u00fc") is False
+
 
 class _Canary:
     """Pickle payload that would run a command on a vulnerable loader."""
@@ -77,6 +92,15 @@ class _Canary:
     def __reduce__(self: Self) -> tuple:
         """Return the os.system call a vulnerable unpickler would execute."""
         return (os.system, (f"touch {self.marker}",))
+
+
+class _FakeMetadata:
+    """Stand-in for a real metadata object of a non-allowlisted class."""
+
+    def __init__(self: Self) -> None:
+        self.train_indices = [1, 2]
+        self.optimizer = "adam"
+        self.epochs = 5
 
 
 class TestRestrictedUnpickler:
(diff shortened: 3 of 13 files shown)


TheColdIce (changes requested):
Reviewed the full diff and ran the branch locally. The core design call is right and most of the implementation is careful — the framing that `pickle.load` and `exec_module` cannot be removed without removing the product, so the boundary is *who reaches them*, is the correct posture for this tool, and `SECURITY.md` says it honestly instead of pretending the sinks are gone.

Requesting changes on one merge-linkage issue and three small bugs.

## Blocker

**`Fixes #453` will close a CVE-reserved report whose headline recommendation is not implemented.** The first fix line in #453 is "Do not run `pickle.load()` on untrusted uploads"; this PR defers that to PR 3. Merging auto-closes the issue while the sink is still there. Suggest `Refs #453` and closing it when PR 3 lands, or a comment on #453 stating that authentication is the accepted mitigation and the format change is tracked separately. Given this is a public disclosure with a reserved CVE (CAN-2026-2036101) and a reporter who already waited out a disclosure window, the paper trail is worth getting right.

## Bugs

**1. Non-ASCII token causes an unauthenticated 500.** `secrets.compare_digest` raises `TypeError` on non-ASCII `str`. Verified against this branch:

```
GET /jobs?token=%C3%BC                        -> 500
Authorization: Bearer \xfc  (latin-1 encoded) -> 500
```

Any unauthenticated caller can force 500s and traceback noise in the auth path. Guard with `presented.isascii()` before comparing, or encode both sides to bytes.

**2. `_SAFE_NAME` and `_SAFE_SUFFIX` accept a trailing newline.** Python's `$` matches before a trailing `\n`, and these use `.match()` rather than `.fullmatch()`. Verified:

```python
safe_name("model1\n")   # -> ACCEPTED, returns 'model1\n'
safe_suffix("x.pkl\n")  # -> '.pkl\n'
```

Not a traversal, but it breaks the single-clean-path-component invariant the module exists to enforce: `m1` and `m1\n` become two visually identical directories on disk, and since `inspector._try_load` dispatches on `path.suffix`, a `".pkl\n"` suffix matches no branch and the upload becomes uninspectable. `re.fullmatch` or `\Z` fixes both. `TrainParams.name` has the same issue — pydantic v2 `pattern` is search-semantics.

**3. `?token=` is accepted on every protected HTTP route, not just the WebSocket.** `@app.middleware("http")` never sees WebSocket scopes, and the WS handler already does its own `websocket.query_params.get("token")` check — so the HTTP query fallback in `_require_auth` buys nothing, and it costs this:

```
INFO: 127.0.0.1:45054 - "GET /jobs?token=SECRETTOKEN123 HTTP/1.1" 200 OK
```

Verified in uvicorn's access log. Suggest dropping the fallback. The WS route genuinely needs the query parameter, and that residual exposure is worth a line in `SECURITY.md` — it reasons carefully about tracebacks leaking paths but doesn't mention the token reaching the access log.

## Design and docs

- **`RestrictedUnpickler` is bypassed by the next call in the normal flow.** It guards `validate_model_metadata`, but `check_compat` runs `joblib.load` on the same `model_metadata.pkl` a few lines later (`checker.py:47`), as do `worker.py:240` and both sample-data endpoints (`main.py:1228`, `main.py:1260`). All authenticated, so this is consistent with the threat model — but "a canary pickle that provably executes under plain `pickle.load` is inert under the restricted unpickler" holds for one endpoint, not for the flow. Worth stating in `SECURITY.md`.

  The unpickler itself holds up on inspection: allowlist-first `find_class` (the actual fix versus the old `super()`-first version), the `EXT*` extension-registry path also routes through `find_class`, `persistent_load` defaults to raising, and `REDUCE` against a stub is inert.

- **`PROTECTED_PREFIXES = ("/jobs",)` is deny-by-exception.** Any route added later outside `/jobs` is unauthenticated by default. For a module whose whole job is the boundary, consider inverting: authenticate everything, allowlist the static SPA mount.

- **Dead branch:** `path_is_protected`'s `path.startswith(p + "?")` can never fire, since `request.url.path` excludes the query string.

- **Docstrings disagree with the code:** `security.py` says `LEAKPRO_WEBAPP_DATA_ROOTS` defaults to "the job directory plus the LeakPro repo root" (it is only the repo root), and that `LEAKPRO_WEBAPP_ORIGINS` defaults to two origins (the code has four).

- **`TrainParams`** gained bounds on every field except the one invariant that matters: `f_train + f_test` can still exceed 1.

## Tests

11/11 pass locally. Two things to tighten:

- **No `tmp_path` isolation.** Running the file leaves 3 job directories in `webapp_jobs/` at repo root with no cleanup, growing on every run — and `_load_jobs()` picks them up on the next real backend start. `test_traversal_in_model_name_rejected` also asserts `not Path("/tmp/pwned_abs").exists()`, which is flaky on a shared machine. Monkeypatching `JOBS_ROOT` to `tmp_path` handles both.

- **`test_preserves_field_names` never exercises the stub path.** It pickles a plain `dict`, whose class is in the allowlist. The case that matters for `validate_model_metadata` is a `ModelMetadata` *object*, where field names survive only via `_make_stub.__setstate__`. That is the half worth testing.

- On the `importorskip` guard: keeping these out of CI is a reasonable call for optional tooling, but consider phrasing the comment affirmatively — "webapp tests; run locally with the webapp extras installed" rather than "skip where fastapi is not installed, e.g. CI" — so the next contributor reads it as intent rather than as a build workaround.

## What is good

- **Middleware ordering is correct, and subtly so.** The auth middleware is registered first and `CORSMiddleware` second; since `add_middleware` inserts at index 0, CORS ends up outermost, so preflight is handled before auth and 401/403 responses still carry CORS headers.
- `allow_credentials` deliberately omitted with a bearer scheme. Right call.
- `start_audit`'s check-and-set is genuinely atomic (no `await` between the check and the write), and `_load_jobs` already demotes a stale `running` to `failed` on restart, so the claim does not introduce a stuck-job path. I went looking for that and it was already handled.
- The `Path("/a/b") / "/etc"` framing is the correct diagnosis — this was an arbitrary-write primitive, not a traversal, and the docstring says so.
- `security.py` is ruff-clean under the repo's strict config even though CI excludes `leakpro/webapp` from linting.

The blocker and the three bugs are all small. Happy to take the bug fixes as a follow-up commit if that is easier.

fazelehh:
Thanks @TheColdIce — this is exactly the review this PR needed. All four actionable items are addressed in 1b8dcc85; every bug was reproduced before fixing.

**Blocker:** agreed and done — the body now says `Refs #453`, with the issue staying open until the safe-format PR (#457) lands. One factual correction to the premise, though: **CAN-2026-2036101 is not a reserved CVE.** We checked before opening this PR — MITRE's CVE Services API returns `CVE_RECORD_DNE` and NVD returns zero results for CVE-2026-2036101; the `CAN-` prefix was retired ~2005 and the number is far outside the 2026 assignment range. So there is currently no CVE linkage to preserve. The paper-trail argument stands on its own merits regardless, which is why we made the change.

**Bug 1 (non-ASCII token → 500):** reproduced (`GET /jobs?token=ü` → 500). Tokens now compare as UTF-8 bytes; garbage credentials get a 401. Regression test added.

**Bug 2 (trailing newline):** reproduced for `safe_name`/`safe_suffix`; both use `fullmatch` now. One correction: `TrainParams.name` was **not** affected — pydantic v2 patterns compile to Rust regex, where `$` does not match before a trailing newline, and `TrainParams(name="m1\n")` already raised `ValidationError` on this branch. Verified before changing anything; the server-side check now simply matches that behavior.

**Bug 3 (`?token=` on HTTP routes):** reproduced, fallback dropped — header-only for HTTP. The WS route keeps its query parameter out of necessity, and `SECURITY.md` now carries the access-log caveat you suggested.

**Fail-closed auth:** took your suggestion — `PROTECTED_PREFIXES` is gone, replaced by a public allowlist of the SPA paths (`/`, `/index.html`, `/logo.jpg`, `/assets/`); everything else 401s by default. The dead `+ "?"` branch went with it. Caught in the process: the startup banner still referenced the old constant and would have crashed at boot — now covered by a smoke assertion.

Also in the commit: `f_train + f_test <= 1` validator (the train path was clamping it silently), both docstring corrections, test isolation to `tmp_path` (they were indeed leaving directories in `webapp_jobs/` that `_load_jobs` would pick up), the de-flaked absolute-path assertion, a stub-`__setstate__` test with a non-allowlisted class, and the affirmative phrasing on the `importorskip` comment.

On the `RestrictedUnpickler` scope point: agreed it guards one endpoint, not the flow — `SECURITY.md` now says so explicitly. Making the whole flow safe for third-party files is what #457 does (safe formats converted at the upload boundary), which is also where the "don't `pickle.load` untrusted uploads" recommendation from #453 gets its real answer.

17/17 tests pass; ruff clean on the touched files.

TheColdIce (approved):
Nice!

PR #456, opened 2026-09-02
chore(webapp): webapp extras, security tests in CI, deploy hygiene

fazelehh:
Hardening follow-up to #455 (stacked on that branch — merge #455 first; this PR then retargets to `main`). None of these are reachable by an unauthenticated caller once #455 lands, which is why they are split out of the security diff.

- **`[webapp]` extras group** in `pyproject.toml` (`pip install -e ".[webapp]"`) replacing the ad-hoc pip line in the webapp READMEs
- **CI now installs `.[dev,webapp]`** — the security regression tests from #455 actually run in CI instead of skipping on missing fastapi
- **Docker volume fix**: `JOBS_ROOT` is overridable via `LEAKPRO_WEBAPP_JOBS_DIR` and compose points it at the `leakpro_jobs` volume — previously job data landed in the container's writable layer and was lost on recreate
- **nginx**: security headers (`X-Content-Type-Options`, `X-Frame-Options`, `Referrer-Policy`), exact-prefix `/jobs/` match, `client_max_body_size` aligned with the backend upload cap
- **Frontend deps**: removed the unused `xlsx` package (two high-severity advisories, no fix available) and applied `npm audit fix` — `npm audit` is now clean; `tsc` and `vite build` verified

Deliberately deferred (tracked separately): per-connection WebSocket log fan-out (a second client currently steals log lines), Dockerfile non-root user + dependency pinning, and the core-library cleanups outside the webapp.


leakpro/webapp/frontend/nginx.conf
```diff
@@ -3,13 +3,21 @@ server {
...
+    client_max_body_size 2048m;
+
     # SPA fallback
     location / {
         try_files $uri $uri/ /index.html;
     }
 
     # Proxy API + WS to backend
-    location /jobs {
+    location ^~ /jobs/ {
```
Muhaddisabarat, line 36: **Bug: this breaks `POST /jobs` and `GET /jobs` behind the dockerized nginx proxy.**

`location ^~ /jobs/` requires a trailing slash. nginx prefix matching is a literal string match (not path-segment-aware), so the exact URI `/jobs` does **not** match `^~ /jobs/` — it falls through to `location /` instead, which serves `index.html` via `try_files` rather than proxying to the backend.

The frontend calls exactly that bare path for the two most fundamental calls in the app:
```ts
// leakpro/webapp/frontend/src/api.ts:65-66
createJob: () => post<{ job_id: string; status: string; created_at: string }>("/jobs"),
listJobs: () => get<JobListItem[]>("/jobs"),
```
So in the docker-compose deployment — the exact setup this PR hardens — step 1 of the wizard ("create job") and the job list would silently get HTML back instead of JSON and fail. I checked the full file on this branch; there's no other block that catches the bare `/jobs` request, and every other backend route already has the trailing slash (`/jobs/{id}/...`), so only these two calls are affected.

Fix — match the bare path exactly, and keep the prefix match for everything under it (this also preserves the original intent of not matching something like `/jobsxyz`). This can't be expressed as a single-line GitHub suggestion without duplicating the unchanged `proxy_pass`/header lines below (they sit just outside this diff's hunk), so here's the full replacement for the block:

```nginx
    location = /jobs {
        proxy_pass http://backend:8000;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $host;
    }
    location ^~ /jobs/ {
        proxy_pass http://backend:8000;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $host;
    }
```

Muhaddisabarat (commented):
Reviewed the diff end to end (pyproject.toml extras, CI, JOBS_ROOT/docker volume fix, nginx, and the xlsx removal).

- `[webapp]` extras + `pip install .[dev,webapp]` in CI: correct fix — the 11 security regression tests from #455 were being silently skipped in CI before this (`pytest.importorskip("fastapi", ...)`), so they weren't actually enforced.
- `JOBS_ROOT` via `LEAKPRO_WEBAPP_JOBS_DIR` + compose setting it to `/data/jobs`: correct — before this, job data was written under the app's own path inside the container, not the mounted `leakpro_jobs` volume, so it was lost on container recreate.
- `xlsx` removal from `package.json`/`package-lock.json`: verified clean — no remaining imports of `xlsx` anywhere in `frontend/src/`.
- nginx security headers and `client_max_body_size` alignment with the backend's upload cap: no issues.

One blocking issue found in the nginx routing change — see the inline comment on `nginx.conf`. Everything else looks good to me.

fazelehh:
Thanks @Muhaddisabarat — the nginx catch is a real one: the bare `/jobs` (createJob/listJobs) would indeed have fallen through to the SPA and returned HTML in exactly the dockerized setup this PR targets. Fixed in ad6d77e5 with your two-block suggestion (`location = /jobs` + `location ^~ /jobs/`), plus a comment explaining why both blocks exist so the next edit doesn't re-break it.

Also note: with #455 merged, this PR is retargeted to `main`.

leakpro/webapp/frontend/nginx.conf
```diff
@@ -3,13 +3,32 @@ server {
...
     }
 
-    # Proxy API + WS to backend
-    location /jobs {
+    # Proxy API + WS to backend. Two blocks: nginx prefix matching is a
+    # literal string match, so `^~ /jobs/` alone would NOT match the bare
+    # `/jobs` that createJob/listJobs call — it would fall through to the SPA
+    # and return HTML. The exact match catches the bare path; the prefix match
+    # covers everything under it without also matching e.g. /jobsxyz.
+    location = /jobs {
```
Muhaddisabarat, line 24: `client_max_body_size` was raised to 2048m here for large model/dataset uploads, but both `/jobs` proxy blocks still use nginx's default 60s inactivity timeout. On a slow/congested connection a large upload can stall for 60s+ with no bytes transferred and get aborted, even though it's well under the new size cap.

Suggest adding to both `location` blocks:
```
proxy_read_timeout 600s;
proxy_send_timeout 600s;
client_body_timeout 600s;
```

Muhaddisabarat:
**leakpro/webapp/Dockerfile.backend** (not touched by this PR's diff, so noting here instead of inline):

This PR adds a clean `webapp` extras group in `pyproject.toml` (`fastapi`, `uvicorn[standard]`, `python-multipart`, `httpx`) and switches CI/README to use it, but `Dockerfile.backend` still hand-lists the same packages instead of adopting `.[webapp]`:

```
RUN pip install --no-cache-dir -e ".[dev]" \
    fastapi uvicorn[standard] python-multipart pyyaml
```

That leaves two dependency lists that can silently drift apart over time — in the file that defines the actual deployed image. Suggest:

```
RUN pip install --no-cache-dir -e ".[dev,webapp]"
```

`pyyaml" is already a base dependency in `pyproject.toml`, so nothing is lost by dropping it from this line.

fazelehh:
Both new points addressed in 711666c4:

- **Proxy timeouts**: added `proxy_read_timeout`/`proxy_send_timeout`/`client_body_timeout 600s` to both `/jobs` blocks, with a comment tying them to the upload size cap so they travel together.
- **Dockerfile drift**: agreed — that was the one place still hand-listing the webapp deps right after this PR created the extras group for exactly that reason. Now `pip install --no-cache-dir -e ".[dev,webapp]"`; verified `pyyaml` is a base dependency in `pyproject.toml` so dropping it from the line loses nothing.

Muhaddisabarat (approved):
Verified both fixes from the earlier review (nginx proxy timeouts, `Dockerfile.backend` using `.[dev,webapp]`) — both look correct. Checked the rest of the diff (CI workflow, README, `JOBS_ROOT` env override, docker-compose, dependency cleanup) and nothing else stands out. CI is green.

PR #457, opened 2026-09-02
feat(webapp): safe dataset formats — .npz and Parquet convert at upload

fazelehh:
Implements the safe-format recommendation from #453. Stacked on #455 (merge that first; this then retargets).

**The gap this closes:** #455's authentication decides *who* may call the server — it does nothing about *what* a file does when loaded. Auditing a model you did not train (model zoo, Kaggle, a vendor checkpoint) is a core LeakPro use case, and a hostile pickled dataset executes on the operator's machine the moment it is inspected.

**How:** new `dataio.py` converts safe formats once, at the trust boundary:

- `.npz` — `data` + `targets` entries, loaded with `allow_pickle=False`
- Parquet / CSV / JSON(L) — column names become `feature_names`, label column auto-detected
- bare `.npy` — data array, `allow_pickle=False`

The upload becomes a **server-generated** `ArrayDataset` pickle, so everything downstream (training path, audit worker, core `_load_population`) keeps consuming a pickle — but of a known class, written by the server. The uploaded bytes are never unpickled by anyone. A safe-format file smuggling pickled objects (object arrays) is rejected with 400, never handed to a pickle fallback.

`ArrayDataset` keeps raw arrays in `.data`/`.targets` (the interface `sample_data`/`sample_image` and core's population indexing already rely on) and normalises in `__getitem__` exactly as the training path did.

**Legacy `.pkl` datasets still work** for files the operator created; the upload UI now warns when a pickle-based file is selected and recommends `.npz`/Parquet for third-party data. `SECURITY.md` replaces the "treat every uploaded file as trusted input" guidance with the honest per-format reality.

**Verification:** 7 new tests (`test_webapp_dataio.py`) — npz and Parquet round-trips through the endpoint and `sample_data`, hostile npz rejected at upload, legacy fallback intact; all 18 webapp tests pass; `tsc` and `vite build` clean.

Refs #453.

### DIFF
--- a/leakpro/webapp/backend/dataio.py
+++ b/leakpro/webapp/backend/dataio.py
@@ -106,6 +106,9 @@ def _from_npz(path: Path) -> tuple[np.ndarray, np.ndarray, None]:
             status_code=400,
             detail=f"No target array found in .npz (looked for {_TARGET_KEYS}; file has {names}).",
         )
+    if data.ndim == 0 or targets.ndim == 0:
+        raise HTTPException(status_code=400,
+                            detail="Expected arrays in the .npz, got a 0-dimensional scalar.")
     if len(targets) != len(data):
         raise HTTPException(status_code=400,
                             detail=f"data has {len(data)} rows but targets has {len(targets)}.")
@@ -121,6 +124,10 @@ def _from_npy(path: Path) -> tuple[np.ndarray, np.ndarray, None]:
             detail="Could not read the .npy file (corrupt, empty, or contains pickled "
                    "objects); or use .npz with data + targets entries.",
         ) from e
+    if data.ndim == 0:
+        # np.load succeeds on a scalar array, but len() below would TypeError.
+        raise HTTPException(status_code=400,
+                            detail="Expected an array, got a scalar .npy file.")
     return data, np.zeros(len(data), dtype=np.int64), None
 
 
--- a/leakpro/webapp/backend/dataio.py
+++ b/leakpro/webapp/backend/dataio.py
@@ -1,0 +1,180 @@
+"""Safe dataset formats for the webapp.
+
+Pickled datasets execute arbitrary code when loaded — unavoidable for the
+legacy ``.pkl`` path, but a model or dataset fetched from a model zoo or a
+Kaggle page is exactly the kind of file a privacy-audit user wants to inspect
+*before* trusting. This module gives them formats that structurally cannot
+execute code:
+
+- ``.npz`` (image / time-series / arrays): ``data`` + ``targets`` entries,
+  loaded with ``allow_pickle=False``.
+- ``.parquet`` / ``.csv`` / ``.json`` / ``.jsonl`` (tabular): column names
+  become ``feature_names``; the label column is auto-detected.
+- ``.npy``: a bare data array (labels default to zeros), ``allow_pickle=False``.
+
+Uploads in these formats are converted once, at the trust boundary, into an
+:class:`ArrayDataset` that the server pickles itself. Everything downstream —
+the training path, the audit worker, LeakPro core's ``_load_population`` —
+keeps consuming a pickle, but it is now a *server-generated* pickle of a known
+class; the user-supplied bytes are never unpickled by anyone.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any, Optional
+
+import joblib
+import numpy as np
+import torch
+from fastapi import HTTPException
+
+from .inspector import _guess_type
+from .models import DataMeta
+
+#: Formats that load without any code-execution risk.
+SAFE_DATA_SUFFIXES = {".npz", ".npy", ".parquet", ".csv", ".json", ".jsonl"}
+
+_DATA_KEYS = ("data", "x", "X", "features", "samples")
+_TARGET_KEYS = ("targets", "y", "Y", "labels", "label")
+_LABEL_COLUMNS = ("label", "target", "y", "class")
+
+#: Filename of the server-generated dataset pickle inside a job directory.
+CONVERTED_NAME = "data_converted.pkl"
+
+
+class ArrayDataset(torch.utils.data.Dataset):
+    """Array-backed dataset with the interface the rest of LeakPro expects.
+
+    ``.data`` and ``.targets`` hold the *raw* arrays (so ``sample_data`` /
+    ``sample_image`` and core's ``population.data[indices]`` see the original
+    values, matching the semantics of the example ``UserDataset`` classes);
+    ``__getitem__`` applies the same normalisation the training path applies:
+    float32, ÷255 for raw pixel data, channel-last → channel-first.
+    """
+
+    def __init__(self, data: np.ndarray, targets: np.ndarray,
+                 feature_names: Optional[list] = None, **_kwargs: Any) -> None:  # noqa: ANN401
+        self.data = data
+        self.targets = targets
+        self.feature_names = feature_names
+
+    def __len__(self) -> int:
+        """Number of samples."""
+        return len(self.data)
+
+    def __getitem__(self, idx: int) -> tuple:
+        """Return one (input, label) pair as normalised tensors."""
+        x = torch.as_tensor(np.asarray(self.data[idx]), dtype=torch.float32)
+        if np.asarray(self.data[idx]).dtype == np.uint8 or x.max() > 1.0:
+            x = x / 255.0
+        if x.ndim == 3 and x.shape[-1] in (1, 3, 4) and x.shape[0] > 4:
+            x = x.permute(2, 0, 1).contiguous()
+        y = torch.as_tensor(self.targets[idx], dtype=torch.long)
+        return x, y
+
+
+def _pick(mapping: Any, keys: tuple) -> Optional[np.ndarray]:  # noqa: ANN401
+    for key in keys:
+        if key in mapping:
+            return np.asarray(mapping[key])
+    return None
+
+
+def _from_npz(path: Path) -> tuple[np.ndarray, np.ndarray, None]:
+    try:
+        with np.load(path, allow_pickle=False) as npz:
+            data = _pick(npz, _DATA_KEYS)
+            targets = _pick(npz, _TARGET_KEYS)
+            names = list(npz.files)
+    except ValueError as e:
+        # Raised for object (pickled) arrays — refuse rather than fall back.
+        raise HTTPException(
+            status_code=400,
+            detail="The .npz contains pickled objects; only plain arrays are accepted.",
+        ) from e
+    if data is None:
+        raise HTTPException(
+            status_code=400,
+            detail=f"No data array found in .npz (looked for {_DATA_KEYS}; file has {names}).",
+        )
+    if targets is None:
+        raise HTTPException(
+            status_code=400,
+            detail=f"No target array found in .npz (looked for {_TARGET_KEYS}; file has {names}).",
+        )
+    if len(targets) != len(data):
+        raise HTTPException(status_code=400,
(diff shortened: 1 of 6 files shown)


leakpro/webapp/backend/dataio.py
```diff
@@ -0,0 +1,180 @@
...
+        with np.load(path, allow_pickle=False) as npz:
+            data = _pick(npz, _DATA_KEYS)
+            targets = _pick(npz, _TARGET_KEYS)
+            names = list(npz.files)
+    except ValueError as e:
+        # Raised for object (pickled) arrays — refuse rather than fall back.
+        raise HTTPException(
+            status_code=400,
+            detail="The .npz contains pickled objects; only plain arrays are accepted.",
+        ) from e
```
Muhaddisabarat, line 98: **Only `ValueError` is caught — a corrupt/truncated `.npz` raises `zipfile.BadZipFile`, which isn't a `ValueError`.**

Reproduced: truncating a valid `.npz` mid-file (breaking the zip central directory) raises `zipfile.BadZipFile: File is not a zip file` from `np.load()`. Since `upload_data()` has no surrounding try/except, this becomes an unhandled 500 on the main upload endpoint instead of the 400 this module promises.

```python
# add near the top of the file: import zipfile

def _from_npz(path: Path) -> tuple[np.ndarray, np.ndarray, None]:
    try:
        with np.load(path, allow_pickle=False) as npz:
            data = _pick(npz, _DATA_KEYS)
            targets = _pick(npz, _TARGET_KEYS)
            names = list(npz.files)
    except (ValueError, zipfile.BadZipFile, OSError, EOFError) as e:
        # ValueError: pickled/object arrays. BadZipFile/OSError/EOFError: a
        # truncated or otherwise corrupt upload, not necessarily malicious --
        # either way this boundary must reject with 400, never raise raw.
        raise HTTPException(
            status_code=400,
            detail="Could not read the .npz file (corrupt, or contains pickled objects).",
        ) from e
```

leakpro/webapp/backend/dataio.py
```diff
@@ -0,0 +1,180 @@
...
+def _from_npy(path: Path) -> tuple[np.ndarray, np.ndarray, None]:
+    try:
+        data = np.load(path, allow_pickle=False)
+    except ValueError as e:
+        raise HTTPException(
+            status_code=400,
+            detail="The .npy contains pickled objects; only plain arrays are accepted "
+                   "(or use .npz with data + targets entries).",
+        ) from e
+    return data, np.zeros(len(data), dtype=np.int64), None
```
Muhaddisabarat, line 131: **Same gap here — an empty `.npy` raises `EOFError`, not `ValueError`.**

Reproduced: `np.load()` on a zero-byte `.npy` raises `EOFError: No data left in file`, uncaught by this except clause.

```python
def _from_npy(path: Path) -> tuple[np.ndarray, np.ndarray, None]:
    try:
        data = np.load(path, allow_pickle=False)
    except (ValueError, OSError, EOFError) as e:
        raise HTTPException(
            status_code=400,
            detail="Could not read the .npy file (corrupt, empty, or contains pickled objects) "
                   "(or use .npz with data + targets entries).",
        ) from e
    return data, np.zeros(len(data), dtype=np.int64), None
```

leakpro/webapp/backend/dataio.py
```diff
@@ -0,0 +1,180 @@
...
+        df = pd.read_parquet(path)
+    elif suffix == ".csv":
+        df = pd.read_csv(path)
+    else:
+        df = pd.read_json(path, lines=(suffix == ".jsonl"))
+
+    label_col = next((c for c in _LABEL_COLUMNS if c in df.columns), None)
+    if label_col is None:
+        label_col = df.columns[-1]
+    features = df.drop(columns=[label_col])
```
Muhaddisabarat, line 153: **`pd.read_parquet`/`pd.read_csv`/`pd.read_json` (and the `df.columns[-1]` fallback) aren't guarded at all** — only the later `.to_numpy()` conversion is (lines 138-145). A malformed Parquet/CSV/JSON file, or one with zero columns, raises whatever pandas/pyarrow/the JSON decoder raises natively and propagates as an unhandled 500 via `upload_data()`.

```python
def _from_table(path: Path) -> tuple[np.ndarray, np.ndarray, list]:
    import pandas as pd
    suffix = path.suffix.lower()
    try:
        if suffix == ".parquet":
            df = pd.read_parquet(path)
        elif suffix == ".csv":
            df = pd.read_csv(path)
        else:
            df = pd.read_json(path, lines=(suffix == ".jsonl"))
    except Exception as e:
        raise HTTPException(status_code=400, detail=f"Could not parse {suffix} file: {e}") from e

    if df.shape[1] == 0:
        raise HTTPException(status_code=400, detail="File has no columns.")
    label_col = next((c for c in _LABEL_COLUMNS if c in df.columns), None)
    if label_col is None:
        label_col = df.columns[-1]
    features = df.drop(columns=[label_col])
```

Muhaddisabarat (commented):
Reviewed `dataio.py`, the `main.py`/`security.py` wiring, the frontend warning UI, and `SECURITY.md`.

The core design is sound and I verified the central claim holds: a `.npz` with a pickled/object array (`test_npz_with_pickled_objects_rejected`) is correctly caught in `_from_npz` and turned into a 400 — it never reaches a pickle loader. `set_data_path` also correctly re-raises `HTTPException` before its generic catch-all, so `convert_upload`'s specific error messages survive there. The `ArrayDataset` normalisation, the frontend pickle warning, and the `SECURITY.md` rewrite all look correct.

One class of issue found, in three places — see the inline comments on `dataio.py`. Summary: the exception handling in `_from_npz`/`_from_npy`/`_from_table` only anticipates the *adversarial* failure mode (pickled objects), not a merely *corrupt* upload (truncated file, empty file, malformed table). I reproduced both gaps locally:

```
truncated .npz  -> zipfile.BadZipFile  (not a ValueError)
empty .npy      -> EOFError            (not a ValueError)
```

Because `upload_data()` in `main.py` has no try/except around `convert_upload()` (unlike `set_data_path`, which has a catch-all), any of these produce an unhandled 500 on the primary upload flow instead of the clean 400 this module's own docstring promises ("Malformed safe-format files raise `HTTPException(400)` ... they are never handed to a pickle-based fallback"). None of the 7 new tests exercise a structurally corrupt file, only well-formed containers with "wrong" semantic content, so this wasn't caught. Not a security hole on its own (no info leak, still behind auth), but it breaks the guarantee this PR is specifically about, on the more likely real-world case (a bad/partial upload) rather than the adversarial one.

fazelehh:
Thanks @Muhaddisabarat — agreed on all three, and the framing is right: the module promised 400 on malformed input but only delivered it for the adversarial case, not the far more likely corrupt/truncated upload. Fixed in b4939f3e, essentially as you suggested:

- `_from_npz` catches `(ValueError, zipfile.BadZipFile, OSError, EOFError)`
- `_from_npy` catches `(ValueError, OSError, EOFError)`
- `_from_table` guards the `read_parquet`/`read_csv`/`read_json` call itself and rejects zero-column frames

Reproduced each failure before fixing (truncated `.npz` → `BadZipFile`, empty `.npy` → `EOFError`, garbage parquet → pyarrow error), and added the missing test class: structurally corrupt files at the unit level plus a truncated `.npz` through the upload endpoint — the 500-path that none of the original 7 tests exercised. 26/26 webapp tests pass.

Also note: with #455 merged, this PR is retargeted to `main`.

leakpro/webapp/backend/dataio.py
```diff
@@ -0,0 +1,189 @@
...
+def _from_npy(path: Path) -> tuple[np.ndarray, np.ndarray, None]:
+    try:
+        data = np.load(path, allow_pickle=False)
+    except (ValueError, OSError, EOFError) as e:
+        raise HTTPException(
+            status_code=400,
+            detail="Could not read the .npy file (corrupt, empty, or contains pickled "
+                   "objects); or use .npz with data + targets entries.",
+        ) from e
+    return data, np.zeros(len(data), dtype=np.int64), None
```
Muhaddisabarat, line 131: A scalar (0-dimensional) `.npy` array slips past the `except (ValueError, OSError, EOFError)` above and crashes here with an uncaught `TypeError`, not a clean 400.

Reproduced:
```python
import numpy as np
np.save('x.npy', np.array(5))
d = np.load('x.npy', allow_pickle=False)  # loads fine, ndim == 0
len(d)  # TypeError: len() of unsized object
```
`np.load()` succeeds (it's a structurally valid file), so the `try/except` around it never fires — the `TypeError` comes from `np.zeros(len(data), ...)` on the next line.

Same root cause applies in `_from_npz` above (line 109's `len(targets) != len(data)`) if either array is 0-d.

Suggest rejecting 0-d arrays explicitly before the `len()` calls, e.g.:
```python
if data.ndim == 0:
    raise HTTPException(status_code=400, detail="Expected an array, got a scalar .npy file.")
```
(and similarly for `targets` in `_from_npz`).

leakpro/webapp/backend/main.py
```diff
@@ -473,7 +474,14 @@ async def upload_data(job_id: str, file: UploadFile) -> DataMeta:
     job = _get_job(job_id)
     dest = _job_dir(job_id) / f"data{safe_suffix(file.filename)}"
     save_upload(file.file, dest)
-    meta = inspect(dest)
+    # Safe formats (.npz/.parquet/.csv/...) are converted here, at the trust
+    # boundary, into a server-generated pickle — the uploaded bytes are never
+    # unpickled. Legacy pickle formats fall through to inspect().
+    converted = convert_upload(dest, _job_dir(job_id))
```
Muhaddisabarat: Unlike `set_data_path` below (which this PR just wrapped in `try: ... except HTTPException: raise / except Exception: -> 400`), `upload_data` has no try/except around `convert_upload` at all. Any non-`HTTPException` failure inside `dataio.py` (e.g. the 0-d array `TypeError` noted on `dataio.py:124`, or anything similar in the future) becomes a raw, unhandled 500 on the primary upload endpoint — the same "promised 400, delivered 500" gap this PR's last commit fixed elsewhere, just via a different trigger.

Suggest mirroring `set_data_path`'s guard here:
```python
try:
    converted = convert_upload(dest, _job_dir(job_id))
except HTTPException:
    raise
except Exception as e:
    raise HTTPException(status_code=400, detail=safe_detail(e, "Failed to convert upload")) from e
```

fazelehh:
Both addressed in 2b52a349 — and the scalar catch is a good one: `np.load` succeeds on a 0-d array so the corrupt-file except never fires, and the `TypeError` came from the `len()` a line later. Reproduced before fixing.

- `_from_npy` and `_from_npz` now reject 0-dimensional arrays explicitly with 400 (both `data` and `targets` in the npz case, covering the `len(targets) != len(data)` line you pointed at).
- `upload_data` now mirrors `set_data_path`'s guard: `HTTPException` passes through, anything unforeseen is logged server-side and returned as a 400 via `safe_detail` — so the next `dataio` edge case degrades to a clean 400 instead of a 500.

Tests: scalar `.npy` and 0-d `.npz` at the unit level, plus a monkeypatched converter raising `RuntimeError` through the endpoint, asserting 400 and that the internal message doesn't leak into the response. 28/28 webapp tests pass.

Muhaddisabarat (approved):
All 5 outstanding review comments are resolved:
- The 3 error-handling gaps (`BadZipFile`, `EOFError`, unguarded table parsing) fixed in `b4939f3e`.
- The 0-d scalar array crash and the unguarded `convert_upload` call in `upload_data` fixed in `2b52a349`, both with new regression tests (`test_scalar_arrays_rejected`, `test_unexpected_converter_error_becomes_400`).

CI is green. Looks good to merge.

PR #460, opened 2026-09-11
refactor(lira): standardize output variable names

### LINKED ISSUE #444: Standardize naming for target outputs and attack signals across MIA attacks
Different MIA implementations currently use different names for similar values. Examples include:

- base: logits_target
- attack_p: attack_signal and audit_signal
- loss_trajectory: target
- LiRA and MS-LiRA: taget_model_logitsm, target_signals, shadow_models_signals, and sample_target_signals

Use the same naming pattern across all similar MIA attacks. 
The names should make it clear whether a variable contains the model’s original output or a signal calculated from that output. 
For example, use `target_outputs` and `shadow_outputs` or `target_model_outputs`  and `shadow_model_outputs`  for original model outputs.

It might be better with `outputs` than `logits`, because LeakPro also supports regression and forecasting models, which do not necessarily produce logits.

NiteeshGowda72:
## Summary

Standardize local variable names for raw model outputs in the LiRA attack.

## Changes

- Rename `target_logits` to `target_outputs`
- Rename `shadow_logits` to `shadow_outputs`
- Keep `target_signals` and `shadow_models_signals` unchanged
- Preserve the existing `load_logits()` API and attack behavior

## Testing

- `python3 -m py_compile leakpro/attacks/mia_attacks/lira.py`
- `python -m pytest leakpro/tests/mia_attacks/attacks/test_lira.py -v`
  - 7 passed
- `python -m pytest leakpro/tests/mia_attacks/attacks/test_multi_signal_lira.py -v`
  - 16 passed

Closes #444

### DIFF
--- a/leakpro/attacks/mia_attacks/base.py
+++ b/leakpro/attacks/mia_attacks/base.py
@@ -135,13 +135,13 @@ class AttackBASE(AbstractMIA):
         ground_truth_indices = dataloader.dataset.targets.numpy()
         assert n_points == len(ground_truth_indices), "Number of points and labels must be the same"
         logger.info(f"Scoring {n_points} points with BASE attack")
-        logits_target = np.array(ModelLogits()([self.target_model], None, None, dataloader)).squeeze(axis=0)
-        log_conf_target = self._log_conf(logits_target, ground_truth_indices)
+        target_outputs = np.array(ModelLogits()([self.target_model], None, None, dataloader)).squeeze(axis=0)
+        log_conf_target = self._log_conf(target_outputs, ground_truth_indices)
         # run points through shadow models and collect the log confidence values
-        logits_sm = []
+        shadow_outputs = []
         for m in tqdm(self.shadow_models, desc="Scoring with shadow models"):
-            logits_sm.append( np.array(ModelLogits()([m], self.handler, None, dataloader)).squeeze(axis=0) )
-        log_conf_shadow_models = np.array([self._log_conf(x, ground_truth_indices) for x in logits_sm])
+            shadow_outputs.append( np.array(ModelLogits()([m], self.handler, None, dataloader)).squeeze(axis=0) )
+        log_conf_shadow_models = np.array([self._log_conf(x, ground_truth_indices) for x in shadow_outputs])
 
         if self.online is True:
             threshold = logsumexp(log_conf_shadow_models, axis=0) - np.log(self.num_shadow_models)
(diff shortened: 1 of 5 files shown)


fazelehh (changes requested):
These changes only fix lira, but the issue is about all attacks.

NiteeshGowda72:
Thanks for pointing that out. I initially scoped the change to LiRA based on the examples in the issue, but I understand now that the naming standardization should cover all relevant MIA attacks. I’ll update the other affected attack implementations as well and push the changes to this PR.

NiteeshGowda72:
Hi, I have been reviewing the MIA implementations to standardize the naming consistently across the attacks. I have a clarification regarding QMIA, OSLO, and YOQO:

In QMIA, target_logits is populated from the configured self.signal(...), so it appears to represent a derived signal rather than raw model outputs.
In OSLO and YOQO, lprime initially contains raw target-model outputs and is then converted/overwritten into predictions.

Should these cases also be refactored as part of #444? In particular, should QMIA use a signal-oriented name, and should OSLO/YOQO separate the raw target_outputs from the resulting target_predictions?

I want to make sure the naming follows the intended convention before making the changes.

NiteeshGowda72:
I’ve implemented the output naming standardization in the following files:

base.py
loss_trajectory.py
multi_signal_lira.py
rmia.py

The changes standardize names for raw model outputs while keeping calculated attack signals and existing APIs unchanged. I also ran the relevant tests, and all 46 tests passed.

I’ve intentionally left the following three implementations unchanged for now:

qmia.py — the existing target_logits values are produced through the configured signal, so I’d like to confirm whether these should be treated as attack signals rather than raw model outputs.
oslo.py — lprime is used in a way that appears to mix the model output and subsequent predictions, so I’d like clarification on the intended naming distinction.
yoqo.py — similarly, lprime is used for model output/prediction-related values, and I’d like to confirm the intended naming convention before changing it.

I left these three untouched pending clarification to avoid introducing a semantic change based on assumptions.

NiteeshGowda72:
Hi @fazelehh , the changes are ready for review. Could you please take a look when you have time? Thanks!

NiteeshGowda72:
Hi @fazelehh, I’ve addressed the requested changes and pushed the updated implementation. The remaining QMIA/OSLO/YOQO cases are left unchanged pending clarification as noted above. The PR is ready for review. Thank you!

PR #466, opened 2026-09-21
Data extraction attacks generativ

henrikfo:
# Description
Addition of Extraction attacks on generative models. Carlini 2023 and SIDE 2026

## Summary of changes
Adds Carlini and SIDE extraction attacks with shared base classes, a CIFAR-10 diffusion example, and regression tests. Includes stabilized SIDE guidance.

## How Has This Been Tested?
Notebook run and attack results looks promising.
~140 tests is intruduced for the extraction attack and related misc.


leakpro/attacks/extraction_attacks/utils_generative/image_metrics.py
```diff
@@ -0,0 +1,321 @@
...
+        raise ValueError("reference_neighbor_means must contain one value per reference image.")
+    if not torch.isfinite(reference_neighbor_means).all() or torch.any(reference_neighbor_means < 0):
+        raise ValueError("reference_neighbor_means must be finite and non-negative.")
+    nearest_indices, nearest_distances = nearest_reference(
+        candidates,
+        references,
+        block_size=block_size,
+        device=device,
+    )
+    neighbor_means = reference_neighbor_means.detach().cpu()[nearest_indices]
```
fazelehh: **The Carlini adaptive ratio divides by the mean distance from the nearest *reference* to its 50 closest references, counting itself at distance 0. Carlini et al. Section 5.1 defines S_x̂ as the n training images closest to the *generated* image x̂ (Fig. 7 shows the distances from a generated image to all training images).**

Failure scenario: Take a generation x̂ that is novel but sits near a training image x in a dense region, such as a blue-sky cluster. The paper's denominator is x̂'s own neighbourhood, which is close too, so the ratio stays near 1/alpha and x̂ is rejected. This code uses x's neighbourhood instead, and one of its 50 terms is 0 because x is its own neighbour, so the denominator shrinks by about 1/50 and the ratio lands in a different place. The attack then flags or misses candidates on a different basis than the paper, and the extracted-image count it reports cannot be compared with the published CIFAR-10 result. The docstring also says this is the Section 5.1 ratio.

_(automated review, recall mode)_

leakpro/attacks/extraction_attacks/utils_generative/carlini_graph.py
```diff
@@ -0,0 +1,83 @@
...
+    for row_index in range(adjacency.shape[0]):
+        mask = 0
+        for column_index, connected in enumerate(adjacency[row_index].tolist()):
+            if connected and column_index != row_index:
+                mask |= 1 << column_index
+        masks.append(mask)
+    return masks
+
+
+def maximum_clique(adjacency: Tensor) -> list[int]:
```
fazelehh, line 36: **`maximum_clique` runs an exact branch-and-bound search with only a size bound (no colouring). Its prune uses `<`, so it also explores branches that can only tie the best clique. On dense graphs its run time grows exponentially, while the paper only needs to know whether some clique has at least min_clique_size members.**

Failure scenario: The default is 500 generations per condition, and tiled_l2_threshold is uncalibrated, so a loose threshold gives a dense graph. Measured on this branch with random graphs at edge density 0.5: n=140 takes 0.37 s, and n=250 does not finish within 4.5 minutes. A single generic prompt at n=500 therefore hangs the Carlini audit indefinitely.

_(automated review, recall mode)_

leakpro/reporting/extraction_result.py
```diff
@@ -0,0 +1,258 @@
...
+        result_dir = results_dir / self.id
+        data_dir = output_root / "data_objects"
+        data_path = data_dir / f"{self.id}.json"
+        claims_dir = output_root / ".extraction_result_claims"
+        results_dir.mkdir(parents=True, exist_ok=True)
+        data_dir.mkdir(parents=True, exist_ok=True)
+        claims_dir.mkdir(parents=True, exist_ok=True)
+        claim = claims_dir / self.id
+        try:
+            claim.mkdir()
```
fazelehh: **The `.extraction_result_claims/<id>` lock directory is removed only in a Python `finally`, and nothing detects a stale lock. If the process is SIGKILLed or runs out of memory during save, the claim stays forever. Scheduler preflight does not check for claims.**

Failure scenario: An OOM kill hits while np.savez_compressed is writing a large candidate batch. From then on, every run with the same config fails in save() with 'already being saved', even with overwrite_results=true. Because preflight passes, the failure only appears after the full generation run (for example 1M samples) has finished and its compute is lost.

_(automated review, recall mode)_

leakpro/attacks/extraction_attacks/side.py
```diff
@@ -0,0 +1,577 @@
...
+            cohesion = float((normalized_features[member_mask] @ normalized_centroids[cluster_index]).mean())
+            if cohesion >= self.config.cohesion_threshold:
+                retained_centroids.append(raw_centroids[cluster_index])
+                retained_cohesion.append(cohesion)
+        if len(retained_centroids) < 2:
+            raise ValueError(
+                "Fewer than two clusters passed cohesion/min-size filtering; lower cohesion_threshold or min_cluster_size."
+            )
+        centroids = torch.stack(retained_centroids)
+        reassigned = torch.cat(
```
fazelehh, line 279: **After the cohesion/min-size filter drops clusters, every synthetic sample is reassigned to its nearest *retained* centroid, including the samples from the rejected clusters. The classifier is then trained on those samples too, which undoes the purpose of the filter.**

Failure scenario: With clusters=100, suppose 70 clusters fail cohesion_threshold. The roughly 70% of synthetic samples in them are relabelled to whichever retained centroid is nearest and used as positive examples for that surrogate class. The time-dependent classifier learns diffuse classes, the guidance stops pointing at the cohesive (likely memorised) modes, and SIDE under-reports extraction. Dropping the non-retained samples is presumably what Algorithm 1 intends; this point is from memory of the paper [prior], because the SIDE paper is not on disk.

_(automated review, recall mode)_

leakpro/leakpro.py
```diff
@@ -14,21 +16,16 @@
 from torch.utils.data import DataLoader
 
 from leakpro.attacks.attack_scheduler import AttackScheduler
+from leakpro.input_handler.abstract_extraction_input_handler import AbstractExtractionInputHandler
```
fazelehh: **leakpro.py now imports AbstractExtractionInputHandler at module level. That module imports extraction_attacks.protocols, which runs the extraction_attacks package __init__, which in turn imports carlini, side and sklearn.cluster. Every MIA/GIA/MINV user therefore pays for the extraction stack on import, which contradicts the lazy imports this same PR adds for ImageAugmentor and ReportHandler.**

Failure scenario: Verified on the PR head: after `import leakpro.leakpro`, sys.modules contains both 'leakpro.attacks.extraction_attacks.side' and 'sklearn.cluster'. Any import error or slow start in the extraction stack now breaks or slows plain MIA audits. Fix at the root: make extraction_attacks/__init__.py lazy, or move protocols out of the attack package, and import ExtractionHandler only inside the extraction branch.

_(automated review, recall mode)_

leakpro/attacks/extraction_attacks/utils_generative/__init__.py
```diff
@@ -0,0 +1,309 @@
...
+
+def require_authorized(authorized_audit: bool) -> None:
+    """Fail closed unless the caller confirms audit authorization."""
+    if not authorized_audit:
+        raise PermissionError(
+            "authorized_audit must be true. Run extraction only on models and data you are authorized to audit."
+        )
+
+
+def resolve_device(requested: str) -> torch.device:
```
fazelehh, line 60: **`resolve_device` re-implements device selection that already lives in leakpro/utils/device.py::get_device. The copy ignores LEAKPRO_DEVICE and rejects 'hpu'.**

Failure scenario: On a Gaudi host with LEAKPRO_DEVICE=hpu, MIA runs on the HPU while compute_device='auto' for SIDE silently picks CPU or CUDA, and an explicit 'hpu' raises 'Unsupported extraction device type'. There are now two device policies that can disagree within one audit, and future fixes to get_device will not reach extraction.

_(automated review, recall mode)_

examples/extraction/cifar10/cifar10_model.py
```diff
@@ -0,0 +1,485 @@
...
+    """Resolve the device name from ``train_config.yaml``."""
+    if requested == "auto":
+        return torch.device("cuda" if torch.cuda.is_available() else "cpu")
+    device = torch.device(requested)
+    if device.type not in ("cpu", "cuda"):
+        raise ValueError("The official Improved DDPM backend supports CPU and CUDA; use CPU on a Mac.")
+    return device
+
+
+def seed_everything(seed: int) -> None:
```
fazelehh: **The example defines its own `seed_everything` (and a `select_device` at line 80) even though LeakPro already has them. This violates /home/fazeleh/LeakPro/CLAUDE.md, which lists 'seeding every RNG | leakpro/utils/seed.py::seed_everything' and says 'Before writing any helper: grep for it'.**

Failure scenario: The local copy does not set cudnn.deterministic or benchmark=False and does not seed HPU, unlike leakpro.utils.seed.seed_everything. Training the example target is therefore less reproducible than the rest of LeakPro, and the two copies will drift apart.

_(automated review, recall mode)_

leakpro/attacks/extraction_attacks/attack_factory_extraction.py
```diff
@@ -0,0 +1,93 @@
...
+        handler: ExtractionHandler,
+    ) -> AbstractExtraction:
+        """Instantiate a supported extraction attack."""
+        config = cls.validate_config(name, attack_config)
+        if name == "carlini_diffusion":
+            if not isinstance(config, CarliniConfig):
+                raise RuntimeError("Carlini configuration dispatch failed.")
+            conditions = normalize_conditions(handler.get_extraction_conditions())
+            reference_images = handler.get_extraction_reference_images()
+            audit_hash = extraction_audit_hash(
```
fazelehh: **For each attack, create_attack separately calls handler.get_extraction_reference_images() and get_diffusion_adapter() and runs extraction_audit_hash, which SHA-256s every byte of the reference set. Two attacks over the same references therefore hash the whole dataset twice, before any attack starts.**

Failure scenario: The audit.yaml ships with both carlini_diffusion and side. With the full CIFAR-10 training set as float32 references (about 614 MB), the dataset is streamed and hashed twice in blocks of 16 images during LeakPro(...) construction. A provider whose getters load from disk or build the model also loads everything twice. Compute the reference digest once per handler and cache it.

_(automated review, recall mode)_

leakpro/attacks/extraction_attacks/side.py
```diff
@@ -0,0 +1,577 @@
...
+        metrics: dict[str, Any] = {
+            "images_generated": images.shape[0],
+            "retained_clusters": self.cluster_centroids.shape[0],
+            "cluster_cohesion": self.cluster_cohesion,
+            "classifier_epoch_losses": self.training_history,
+            "guidance_scale": self.config.guidance_scale,
+        }
+        if self._references_metric is not None:
+            images_for_l2 = self._reference_metric_images(images, metric_images)
+            nearest_indices, nearest_distances = nearest_reference(
```
fazelehh, line 507: **With image_range=minus_one_one, SIDE computes nearest_reference_distance and its L2 bands in raw [-1,1] coordinates, while Carlini converts to [0,1] before computing the same field. Within one audit, the same CandidateRecord field and the same threshold numbers therefore differ in scale by a factor of 2 between attacks.**

Failure scenario: The shipped audit.yaml uses minus_one_one for both attacks. Carlini sets verified from nearest_reference_distance <= 0.15 on the [0,1] scale, while the SIDE record for an identical image reports twice that distance. Anyone comparing the two results, or reusing Carlini's 0.15 for SIDE's l2_bands, gets extraction counts that are wrong by a factor of 2 in distance. SIDE also never sets `verified`.

_(automated review, recall mode)_

leakpro/attacks/extraction_attacks/configs.py
```diff
@@ -0,0 +1,134 @@
...
+        """Reject inverted distance bands."""
+        if self.lower > self.upper:
+            raise ValueError("SimilarityBand.lower must not exceed upper.")
+        return self
+
+
+class ExtractionConfig(BaseModel):
+    """Shared runtime and authorization settings."""
+
+    random_seed: int = 42
```
fazelehh, line 32: **Each extraction attack has its own `random_seed` defaulting to 42, and nothing reads audit.random_seed. In addition, Carlini's unconditional sampler and SIDE's synthetic-set sampler both seed batch i with random_seed + i.**

Failure scenario: A user who sets audit.random_seed: 7 and no per-attack seeds still runs with seed 42, and repeated audits intended to use different seeds come out identical. With equal seeds, as in the shipped audit.yaml (2026 for both), SIDE's synthetic set regenerates exactly Carlini's first 4096 samples, which wastes a full sampling pass and correlates the two attacks' results.

_(automated review, recall mode)_

fazelehh:
**[Efficiency / code size]** This PR adds 7,057 lines (~2,860 library, ~3,250 tests, ~950 example). The two algorithms themselves are small (Carlini: sample → tiled L2 → clique → compare to train; SIDE: cluster → train classifier → guided sample); most of the size is surrounding machinery. Please cut:

- **`leakpro/reporting/extraction_result.py` (258 lines)** — a transactional save system (staging, backup, restore, lock dir, rollback, `extraction_result.py:146-256`). Write through the existing reporting path instead; this also removes the stale-lock bug flagged inline.
- **`utils_generative/__init__.py` (309 lines)** — ~140 lines of condition canonicalization + hashing (`condition_hash`, `_canonical_*`, `extraction_audit_hash`, `json_safe`), which also drives the double reference hashing flagged in the factory. Plus `resolve_device`, which duplicates `leakpro/utils/device.py::get_device`.
- **`protocols.py` + `adapters.py` (277 lines)** — a three-level Protocol hierarchy (`ExtractionAdapter` → `SamplingAdapter` → `DiffusionAdapter`) for a single model family. One protocol is enough until a second family exists.
- **`examples/extraction/cifar10/cifar10_model.py` (485 lines)** — training a diffusion target is legitimately needed, but the three `sha256_*` helpers, RNG-state checkpoint/restore, and local `seed_everything`/`select_device` are not (use `leakpro/utils/seed.py` and `leakpro/utils/device.py`).
- **Tests (~3,250 lines, more than the library)** — `test_leakpro_integration.py` alone is 966 lines for 25 tests, yet the suite catches none of the inline findings (e.g. the §5.1 scoring basis, clique blow-up). Fewer, targeted tests against paper behaviour would cover more.

Rough target: ~1,200–1,500 library lines with a comparably sized test suite.

leakpro/attacks/extraction_attacks/utils_generative/image_metrics.py
henrikfo: Fixed

leakpro/reporting/extraction_result.py
henrikfo: fixed

leakpro/leakpro.py
henrikfo: fixed

leakpro/attacks/extraction_attacks/side.py
henrikfo: The reassignment described here is intentional and follows Algorithm 1, I might have missed something but does the pape discards samples from rejected clusters, or does reassignment causes the failure?

